PR #27, opened 2013-03-05
KCC Gamma + GUI

AcidWeb:
Implemented Devernay ideas.
Make GUI look less like whirlpool :-)
GUI is not finished - I will mess with it more after merge.


axu2:
Anyone know where the default gamma values came from?

PR #126, opened 2015-02-04
Restored Manga Cover Database support

OpenNingia:
Since a new MCD is up I decided to add back the automatic cover download.

source:
http://www.reddit.com/r/manga/comments/2sqk7p/manga_cover_database_is_back/

link:
http://mcd.iosphe.re/

The API are different, still simple however.


AcidWeb:
Hello,

Thank you for contribution but this feature is already reimplemented in dev branch.

Issue #129, opened 2015-02-21
Random order of chapters in EPUB output


fsteffek:
Issue similar to #127. walk() does not sort directories, hence chapter names from directories are in random order.

AcidWeb:
I will check all walk().

I'm suprised that nobody else reported this issue and I never have this problem myself.

PR #130, opened 2015-02-21
Add bookmarks from ComicInfo.xml as chapters for EPUB output

fsteffek:
I didn't know how to do 2 pull requests from one branch so I include the fix for #129 

I also don't know much about the GUI, so the option to disable the parsing is on commandline only for now.

Thanks for responding so quickly :+1:

### DIFF
--- a/kcc/comic2ebook.py
+++ b/kcc/comic2ebook.py
@@ -422,9 +422,14 @@ def buildEPUB(path, chapterNames, tomeNumber):
                   "}",
                   ])
     f.close()
+    # Ensure we're sorting dirs, files naturally
+    convert = lambda text: int(text) if text.isdigit() else text
+    alphanum_key = lambda key: [convert(c) for c in split('([0-9]+)', key)]
     for (dirpath, dirnames, filenames) in walk(os.path.join(path, 'OEBPS', 'Images')):
         chapter = False
-        filenames.sort()
+        # Traverse dirs in a sorted manner
+        dirnames.sort(key=lambda name: alphanum_key(name.lower()))
+        filenames.sort(key=lambda name: alphanum_key(name.lower()))
         for afile in filenames:
             filename = getImageFileName(afile)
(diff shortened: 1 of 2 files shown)


AcidWeb:
You tested it with files/directories that contain wide unicode chars?

fsteffek:
No, I haven't tested wide unicode chars. I am not sure why you ask for testing files. Do you think sorting them will be a problem?

AcidWeb:
Well considering that I made hundreds of files with KCC and never had any sorting problem I want to be sure :-)

For now I will merge only that fix. I'm pulling rest to separate branch and check later.

fsteffek:
I see. It does seem weird. I have converted quite a few files myself in Windows Vista and 7. My source directories were mostly flat though, so I haven't actively used the chapter feature. Just to clear any misunderstanding: `filelist` **is sorted**. The issue I found lies in that the chapter list is created from `filelist` before the `filelist` is sorted. The same goes for the already fixed cover-issue.

The thing is, even though this issue exists, it may be noticeable, but it does not affect the reading experience like a crash or an empty page would.

Info on walk():
walk() from scandir says to work exactly like os.walk(). The [python docs](https://docs.python.org/2/library/os.html#os.walk) say that os.walk() uses os.listdir() which returns `The list is in arbitrary order`

Thanks for KCC. :)

fsteffek:
Before merging this feature into master, please be aware that my code only works on conversions that do not split images. The chapter-links are calculated based on the original image index number. If you convert comics with big images chapter links might be shifted. Conversions of web-comics are most affected.

Example: 
```
Bookmark to Chapter:
 img-0 [bookmark 1] => img-0 [chapter 1]
 img-2              => img-2
 img-3 (big image)  => img-3-a
                    => img-3-b [chapter 2]
 img-4 [bookmark 2] => img-4
 img-5              => img-5
```

Making it work with all conversions, deeper changes are required.

AcidWeb:
I will try to fix it.

PR #159, opened 2015-09-22
Fix for combo box issue on retina display

kyuzumaki:
This should fix the issue, the combobox is now set to the default height. Not sure exactly how to compile this into an OSX binary to test it fully.


AcidWeb:
It look extremely bad.

Issue #178, opened 2016-01-03
kindlegen stays open forever on very large conversion (100+ mb)


tamodolo:
That's it. On single chapter kindlegen auto close and KCC continue the job while large files it has to be killed manually for kcc to continue.

AcidWeb:
Heard about this issue. Never found the cause.

It is not connected to input size.

tamodolo:
I was using the binary release and that don't show me the verboses. I'm downloading the py version and I'm setting the environment to run it on windows cmd to see what is happening. More info to come if I'm sucessful.

AcidWeb:
Here is function that start kindlegen worker pool:
https://github.com/ciromattia/kcc/blob/master/kcc/comic2ebook.py#L1253
And worker itself:
https://github.com/ciromattia/kcc/blob/master/kcc/comic2ebook.py#L1223

I would start with printing kindlegen output and checking what is last message before it locks.

tamodolo:
It's kindlegen fault. I added a pull request that terminate kindlegen as soon as mobi creation is finished.

Not a beautiful solution but it works now.

AcidWeb:
Great job. I will merge it later.

PR #180, opened 2016-01-03
kindlegen is now forced to terminate issue #178

tamodolo:
A bug in kindlegen causes it to hang randonly when converting epub to mobi. This happens above 90% of the time leaving kcc waiting forever so this change force kindlegen it to be terminated as soon kcc detect it's done the job.


AcidWeb:
Merged to dev branch.

PR #197, opened 2016-06-17
Select multiple directories at once

DoHe:
This pull request allows to select multiple directories in the "Add directory"  dialog. This is helpful if you have a directory per chapter/issue/... with the corresponding images and want to convert several such directories at once.

### DIFF
--- a/kcc/KCC_gui.py
+++ b/kcc/KCC_gui.py
@@ -461,6 +461,8 @@ class ExtendedSelectDirectoriesFileDialog(QtWidgets.QFileDialog):
 
         self.setFileMode(self.Directory)
         self.setOption(self.ShowDirsOnly, True)
+        if os.name == 'nt':
+            self.setOption(self.DontUseNativeDialog, True)
         for view in self.findChildren((QtWidgets.QListView, QtWidgets.QTreeView)):
             if isinstance(view.model(), QtWidgets.QFileSystemModel):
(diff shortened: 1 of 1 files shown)


AcidWeb:
I'm aware that is possible by creating custom widget. But last time I checked this approach don't work with native widgets - and that cause nightmarish look of this form on most of platforms.

DoHe:
I tested it on Ubuntu 16.04. and it looked exactly the same as before. I have a hard time getting it to run on windows but I can try again. I don't have a mac available so I can't test this.
If it should turn out to look bad on anything but Ubuntu, how would you feel about adding another check and make it available only in Linux? I've seen quite some of similar checks in the code already.

AcidWeb:
OK. Maybe they fixed it. I will check this out.

Thanks.

DoHe:
Did you ever get a chance to test this on something that's not Ubuntu?

AcidWeb:
No. No yet.

DoHe:
I got it to run in Windows. My change actually didn't work there at all, but I added a line to make it work there as well. But as you said that deactivates the native file dialog, though I think it still looks reasonably good. Maybe this could be a setting somewhere? Or if you don't like the look at all, I could leave out the line I added and make it a Linux only feature?

AcidWeb:
I should have time this week to check this out.

AcidWeb:
I tested it on QT 5.7 and as I expected it look bad on most of the platforms (Windows, OS X, some Linux WM). So I'm not interested with this change.

PR #214, opened 2016-11-20
add autoscale option

houcheng:
Instead of fixed 1.5 scale ratio, the autoscale feature uses current page's
image width, and dynamically determine the needed scale ratio. The rendering
effects looks okay and speed is fine in my KPW1.

The generated panel view will have two view ports: (top and bottom).


AcidWeb:
Interesting approach. I will check it out.

houcheng:
Thank you~

PR #221, opened 2017-01-05
Fix unreadable text with dark qt themes

bakatrouble:
Before:

After:


AcidWeb:
Thanks - I will add it in next release.

AcidWeb:
Merged with dev branch.

PR #234, opened 2017-04-12
Adding ComicInfo support for PDF

LitchiTheCat:
Awesome tool, but could you add an option so the meta file ,which is in the same directory as the PDF file, is used for creating the file ? And could you add another option so the user can choose how he wants to format the comic title ?
Thaaanks


AcidWeb:
This is not a pull request.

PR #255, opened 2017-12-12
Adding Dockerfile + instruction on README

saoula:
Adding Dockerfile and the instruction in the README file.
Easy to use and no need to worries about dependencies!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3
+MAINTAINER Michel Saoula <dev@saoula.com>
+LABEL com.kcc.name="Kindle Comic Converter"
+LABEL com.kcc.author="Ciro Mattia Gonano and Paweł Jastrzębski"
+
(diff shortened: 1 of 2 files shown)


AcidWeb:
I have some doubts that KindleGen license allow to install it this way.

ciromattia:
@AcidWeb actually [the license](https://www.amazon.com/gp/feature.html?docId=1000599251) only says

> You may not use the Software over a network

which I read as "you may not centralize a kindlegen installation for distributed use", while using Docker it simply becomes a component of an installation - a container is just a "light" VM, after all.

@saoula I couldn't test the pull as I couldn't find anything on docker hub, am I missing something?

AcidWeb:
We would not use saoula/kcc anyway as we don't control content of it.

saoula:
Too bad Kindlegen cannot be included for distributed use.
I thought you might want to create your own docker hub repo with the Dockerfile that I added.

It looks like a dead end with the kindlegen license so I will close this PR.

AcidWeb:
@ciromattia by pushing ready docker image to DockerHub we redistribute KindleGen binary. Additionally user don't know that he need to accept separate EULA to use it.

PR #266, opened 2018-04-18
Use more standard __version__ rather than PILLOW_VERSION

hugovk:
They're both the same thing:

https://github.com/python-pillow/Pillow/blob/master/src/PIL/__init__.py


AcidWeb:
Merged with dev branch.

PR #303, opened 2019-02-24
Fix No Files Bug

ShinonomeHana:
Change *glob()*  to *os.path.exists()*, to fix the "No matching files found" probem when the filename contains "[]".

For example, there is a comic:
*'[河本ほむら×尚村透] 賭ケグルイ 第10巻.zip'*

Because filename contains *[ ]*, so glob will not get the file correct.


AcidWeb:
Thanks for bug report but there is better way to fix it.

PR #305, opened 2019-02-27
[WIP] support pdf rotate

### LINKED ISSUE #304: pdf rotation handling
I am new to kcc,

my collection are personal scan by ScanSnap iX500, 

here is few page of the pdf for demonstrate the issue,
https://1drv.ms/b/s!ArlB91L9JmXYu1LDVm7-k7Wey7zw

I usually feed the page into the scanner side-way to prevent page jamming/tearing during the process,
which looks fine in pdf format,
but unfortunately it seems `kcc-c2e` doesn't handle the page rotation info in pdf,

after some googling looks likd `pdfinfo` is able to extract the info to tackle this problem

https://stackoverflow.com/questions/12050743/how-to-find-whether-pdf-has-landscape-orientation-or-portrait

AndCycle:
initial attempt to fix #304, review required.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -35,6 +35,10 @@ from multiprocessing import Pool
 from uuid import uuid4
 from slugify import slugify as slugifyExt
 from PIL import Image
+
+from imdirect import imdirect_open
+Image.open = imdirect_open
+
 from subprocess import STDOUT, PIPE
 from psutil import Popen, virtual_memory, disk_usage
 from html import escape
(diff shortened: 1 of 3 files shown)


AcidWeb:
I'm not interested with expanding PDF support.

AndCycle:
I do the research to fix part of the PDF support due to you still have interest in taking care of PDF problem on a issue back to 2013 and some question raised due to PDF along the way, 

there is probably only few things need to be take cared around pdf that produced by scanner,
1. jpg support [common, already support]
2. rotation [happens occasionally, due to certain condition limit by scanning process]
3. tiff support [rare, do exist in the wild]

just left the note here if someone wanna carry on this

guess I will keep using personal cooked script.

PR #328, opened 2019-10-27
Enable Synthetic Spreads for the Kobo Forma.

shavitmichael:
Rotating the device into landscape mode will cause it to display two
pages side by side. This is particularly useful for input comics whose
spreads are already split into two files.

See https://github.com/kobolabs/epub-spec#synthetic-spreads for
additional information on synthetic spreads.

(Tiny drive-by: Add Kobo format to list of profiles for which the AUTO
format is EPUB)

* Open question: Should we enable this for other Kobo devices?
I hope they would fallback gracefully if they don't fully support the spec, but I've excluded them from this commit since I don't have any to test one.


AcidWeb:
Thanks I will investigate this topic during refactoring.
I don't have access to any Kobo hardware so entire Kobo support is dry coded.

darodi:
@shavitmichael 
thanks
merged in v5.6.1

I don't own a Kobo Forma to test though

PR #331, opened 2019-11-12
Add profiles for the Kobo Clara HD and Libra H2O

fbriere:
Note: I omitted the "HD" and "H2O" suffixes from the labels to keep
things simple, since they do not denote a variation of a preexisting
model, as was the case with the Aura.


AcidWeb:
Thank you. I will merge it  during refactoring.

PR #343, opened 2020-02-29
Added --croppingminimum option to set a cropping minimum area ratio

### LINKED ISSUE #342: Option to disable cropping for pages with small contents
It is common for [yonkoma manga](https://en.wikipedia.org/wiki/Yonkoma) volumes to include pages (typically between chapters or at the end) featuring only a small illustration.  kcc will currently (when stretch/upscale is enabled) crop and enlarge such images to fit the whole screen, often resulting in something that frankly doesn't look very good.

This issue has been brought up on the MobileRead forum thread a couple of times (examples [here](https://www.mobileread.com/forums/showpost.php?p=2463744&postcount=78), [here](https://www.mobileread.com/forums/showpost.php?p=3253943&postcount=845), [here](https://www.mobileread.com/forums/showpost.php?p=3563364&postcount=1077) and [here](https://www.mobileread.com/forums/showpost.php?p=3693491&postcount=1165)), with suggestions to disable cropping below a given threshold.  @ciromattia was [initially open to the idea](https://www.mobileread.com/forums/showpost.php?p=2464086&postcount=79), but @AcidWeb [later seemed to reject such a proposal](https://www.mobileread.com/forums/showpost.php?p=3694040&postcount=1166) (although I'm not sure if the issue was with the feature itself, or its inclusion in the GUI).

Would you have any objection to such a feature, provided it was optional, and possibly restricted to the CLI?

I can only speak from personal experience, but after implementing this for myself (I'll submit a PR shortly), my results with kcc have changed from "reluctantly accept that many pages will look bad" to "quite pleased that everything looks mighty fine".

fbriere:
Closes #342

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -557,9 +557,9 @@ def imgFileProcessing(work):
         for i in workImg.payload:
             img = image.ComicPage(opt, *i)
             if opt.cropping == 2 and not opt.webtoon:
-                img.cropPageNumber(opt.croppingp)
+                img.cropPageNumber(opt.croppingp, opt.croppingm)
             if opt.cropping > 0 and not opt.webtoon:
-                img.cropMargin(opt.croppingp)
+                img.cropMargin(opt.croppingp, opt.croppingm)
             img.autocontrastImage()
             img.resizeImage()
             if opt.forcepng and not opt.forcecolor:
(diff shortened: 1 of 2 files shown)


fbriere:
I defined the command-line argument as a simple ratio (up to 1.0), but maybe a percentage (up to 100.0) would have been more intuitive?  Feedback welcome.

clach04:
cropping power appears to have same scale so seems reasonable to me :+1:

fbriere:
It just struck me that there's no need to actually choose between the two: both could be supported at the same time, allowing for either 0.85 or 85 as an example.  (In the unlikely situation where you want a minimum of 1% or less, then you would be required to use a ratio as originally planned.)

darodi:
@fbriere 
thanks
merged in v5.6.1

Issue #337, opened 2020-01-11
Feature: Choose chunk size limit


lapwat:
Outputs are divided into 400Mb files as stated here
https://github.com/ciromattia/kcc/wiki/Automatic-output-splitting

Would it be possible to choose this size in a feature? My Kindle Oasis does not accept files bigger than 50Mb for example (even if I just upload it via USB).

I you cannot do it now, can you link me to that value in the code so I can change it manually?

Best

AcidWeb:
You definitively can upload file bigger than 50 mb over USB.

lapwat:
When I upload **Mobi** files on my **Kindle Oasis 2**, the list of files just stops refreshing... I will try with epub.

lapwat:
I have tried again with an epub and still could not make it work.

Here is the how to reproduce:
1. Create an empty folder with a random name (BlueCheese)
1. Copy more than 50Mb of data inside it
1. Generate an epub from this folder (no output split or such options)
1. Check that the epub file is more than 50Mb `ls -lh cp BlueCheese.epub` (63Mb for me)
1. Mount your Kindle `sudo mount /dev/sda /mnt`
1. Copy this folder into /documents on your Kindle `sudo cp BlueCheese.epub /mnt/documents`
1. Unmount your kindle `sudo umount /mnt`
1. Start your Kindle

--> The file does not appear in my books...

I noticed something interesting: the `cp` command is executed instantly, like if the data was not accepted by my Kindle. Hence my request of having the possibility to choose the chunk size limit.

AcidWeb:
Kindles can't read EPUBs.

PR #347, opened 2020-03-06
Added targetsize option

whtsky:
#337

I use kcc on my server to automatically create mobi for mangas and send them to my Kindle via email. And Amazon's Kindle email address only accepts attachments <=50MB.

I did not change the GUI since I don't use them and I know nothing about Qt.


darodi:
- merged in #457

PR #367, opened 2020-07-06
Document CLI tools debian quick install

clach04:
Command Line Tools can be used with pre-built dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,10 @@ After completed conversion, you should find ready file alongside the original in
 Please check [our wiki](https://github.com/ciromattia/kcc/wiki/) for more details.
 
 CLI version of **KCC** is intended for power users. It allows using options that might not be compatible and decrease the quality of output.
+CLI version has reduced dependencies, on Debian based distributions this commands should install all needed dependencies:
+```
+sudo apt-get install python3 p7zip-full python3-pil python3-psutil python3-slugify
+```
 
 ### Standalone `kcc-c2e.py` usage:
 


clach04:
NOTE on my SBC the Linux distro version of PIL is 4.0.0 - so if WebP support is **not** needed, need to disable Pillow 5 check.

I don't think that's worth document in the readme but mentioning for completeness. See https://github.com/ciromattia/kcc/pull/366

    [localhost:~/kcc] # git diff
    diff --git a/kindlecomicconverter/shared.py b/kindlecomicconverter/shared.py
    index 507873b..1435a42 100644
    --- a/kindlecomicconverter/shared.py
    +++ b/kindlecomicconverter/shared.py
    @@ -124,7 +124,8 @@ def dependencyCheck(level):
                 missing.append('python-slugify 1.2.1+')
         try:
             from PIL import __version__ as pillowVersion
    -        if StrictVersion('5.2.0') > StrictVersion(pillowVersion):
    +        #if StrictVersion('5.2.0') > StrictVersion(pillowVersion):
    +        if StrictVersion('4.0.0') > StrictVersion(pillowVersion):
                 missing.append('Pillow 5.2.0+')
         except ImportError:
             missing.append('Pillow 5.2.0+')

clach04:
For more complete diff/commit to revert PIL/Pillow version check see https://github.com/clach04/kcc/commit/590a93da9692eb8d4daa3c3222723e7581dc2b45

darodi:
@clach04 
thanks
merged in v5.6.1

PR #388, opened 2021-02-21
Disable all image transformation : cli and GUI option

StudioEtrange:
Add a --noprocessing cli option and an GUI option to NOT transform image 
it will ignore any selected profil or processing options

In example : usefull when just want to convert some zip files to cbz by just cleaning useless files

Alors add 2 helpers scripts to generate Qt ui related files


StudioEtrange:
Alors add 2 helps scripts to generate Qt ui related files

cellsinterlinkd:
is there a way to make this feature work on 'mobi/azw3' output format?

darodi:
merged in #416

Issue #327, opened 2019-10-21
Major refactoring in progress


AcidWeb:
Hello,
I'm currently working on KCC 6.0 - major internal refactoring plus new GUI.

Branch 5.x is considered EOL and bugs reported for it will be most likely ignored.
There are no plans for another 5.5.x release.

No ETA for the 6.0 release.

KaMyKaSii:
@AcidWeb no ETA but i wonder if you're still working on it? Thank you

clach04:
@AcidWeb is the in-progress v6 branch available anywhere? I just saw this and now realize the PRs I posted are after when this was posted so not likely to get reviewed. An update to the readme would be helpful as there are some PRs from this year as well so I'm not the only one who missed this.

I'm happy to look at code that's not ready yet :)

Thanks for making this tool available!

axu2:
V6 has been dropped. https://github.com/ciromattia/kcc/issues/456#issuecomment-1381798395_

PR #393, opened 2021-05-15
Fixed the skipped/missed images and/or panels

FulyaDemirkan:
This pull request consists of several fixes:

**mergeDirectory()**
**Problem:** If the image width is greater than the targetWidth, app skips that image and omits from converting it. 
**Solution:** if that happens the code resizes the image to the targetWidth while preserving the aspect ratio.

**splitImage()**
**Problem:** If the panel at the bottom of the image is not followed by a solid cropped row then the app doesn't add that panel to the panels array. 
**Solution:** If last the panel is detected but there is no solid cropped row at the bottom of the image, code sets the panel bottom dimension to the heightImg.

**Problem:** If the source image is badly cropped and has a thin black outline it causes app to consider whole image as one panel which ends up with lots of unnecessary empty pages, or tiny panels.
**Solution:** 4 pixels from right and left gets cropped by default to avoid that, It's thin enough to not cause a major problem with the actual image but eliminates the problem with some of the outlines.

**Problem:** Larger sized images (as far as I can see images larger than 1.5-2mb) throws a "DecompressionBombWarning". 
**Solution:** I set the MAX_IMAGE_PIXELS = 1000000000 and put a simple warning.

**makeBook()**
**Problem:** Max height is set to 1024 by default. So, even your device resolution is higher than that images are cropped by 1024. This causes some images that might actually fit to the screen to be cropped to two. 
**Solution:** Max height is now set to device's resolution. (For example for Kindle PW 3+ it's 1448 now.)

I also implemented @pavelzw pull request [for systems with with a lot of RAM](https://github.com/ciromattia/kcc/pull/376)

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1078,10 +1078,7 @@ def makeBook(source, qtgui=None):
     getComicInfo(os.path.join(path, "OEBPS", "Images"), source)
     detectCorruption(os.path.join(path, "OEBPS", "Images"), source)
     if options.webtoon:
-        if image.ProfileData.Profiles[options.profile][1][1] > 1024:
-            y = 1024
-        else:
-            y = image.ProfileData.Profiles[options.profile][1][1]
+        y = image.ProfileData.Profiles[options.profile][1][1]
         comic2panel.main(['-y ' + str(y), '-i', '-m', path], qtgui)
     print("Processing images...")
     if GUI:
@@ -1222,7 +1219,7 @@ def makeMOBI(work, qtgui=None):
         threadNumber = 1
     elif 2 < availableMemory <= 4:
         threadNumber = 2
-    elif 4 < availableMemory <= 8:
+    elif 4 < availableMemory:
         threadNumber = 4
     else:
         threadNumber = None
--- a/kindlecomicconverter/comic2panel.py
+++ b/kindlecomicconverter/comic2panel.py
@@ -57,9 +57,8 @@ def mergeDirectory(work):
         if len(images) > 0:
             targetWidth = max(set(sizes), key=sizes.count)
             for i in images:
-                if i[1] <= targetWidth:
-                    targetHeight += i[2]
-                    imagesValid.append(i[0])
+                targetHeight += i[2]
+                imagesValid.append(i[0])
             # Silently drop directories that contain too many images
             # 131072 = GIMP_MAX_IMAGE_SIZE / 4
             if targetHeight > 131072:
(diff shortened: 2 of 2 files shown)


clach04:
@FulyaDemirkan there are a number of outstanding PRs so I'm not sure when you will get a response to this PR.

The first issue sounds rather scary, I've a number of comics that are wider than the target width of my device and I've not seen the problem you described. Do you have a test case?

For the border of 4 heuristic. Take a look at https://github.com/FooSoft/mangle/blob/9b5254047901f0c8285742eaa084864cd7f63248/mangle/image.py#L183 which can handle arbitrary sizes that might be a safe solution.

Are you able to break these into multiple commits to separate them out? One commit, per bug fix? Again, I'm not sure if/when this will get reviewed so I understand if you don't have time for this.

FulyaDemirkan:
Hi @clach04,

**### For the first issue;**
**1.** `targetWidth = max(set(sizes), key=sizes.count)` sets the targetWidth with the maximum number of occurrences of the width of all the images for the chapter instead of the actual max width. 
**2.** The following code adds the images only with the equal or smaller width to the imagesValid array and completely missing the ones that are larger.
```
if i[1] <= targetWidth:
    targetHeight += i[2]
    imagesValid.append(i[0])
```

So, when I try to convert this chapter, targetWidth is set to 755. Converted chapter ends up missing 14 pages out of 26.
image

For the solution, 
Changing targetWidth to actual max is also not a good solution. In case there is just one image with a huge width then all the smaller ones should be resized which is not efficient. Instead, I am adding all the images to imagesValid.
However, with doing so, then the next step "resizing based on the targetWidth" becomes problematic for larger images. Because
`img = ImageOps.fit(img, (targetWidth, img.size[1]), method=Image.BICUBIC, centering=(0.5, 0.5))`
crops the image.
To fix that problem, instead of using the original height, I am getting the aspect ratio and use that ratio to calculate the targetHeight, Then I use targetWidth and targetHeight to resize the images. Since it works well with the smaller images as well, I applied the same logic to them as well. 
```
if img.size[0] < targetWidth or img.size[0] > targetWidth:
   widthPercent = (targetWidth / float(img.size[0]))
   heightSize = int((float(img.size[1]) * float(widthPercent)))
   img = ImageOps.fit(img, (targetWidth, heightSize), method=Image.BICUBIC, centering=(0.5, 0.5))
```


**### For the second issue:**
Scenario: Panel is detected, but the last tmpImage is not solid (there is no empty space between the drawing and the end of the panel).
In this case, none of the the following if conditions are met and since `while yWork < heightImg` is not also true anymore that panel goes missing.
```
while yWork < heightImg:
   tmpImg = imgProcess.crop([0, yWork, widthImg, yWork + 4])
   solid = detectSolid(tmpImg)
   if not solid and not panelDetected:
      panelDetected = True
      panelY1 = yWork - 2
   if solid and panelDetected:
      panelDetected = False
      panelY2 = yWork + 6
      panels.append((panelY1, panelY2, panelY2 - panelY1))
   yWork += 5
```
So, I added a checkpoint for `if heightImg - yWork <= 5`

I am not sure what you meant by "For the border of 4 heuristic." though. Are you referring to the third problem which crops 4 pixels from the sides?

clach04:
@FulyaDemirkan  thank you for taking the time and effort to make the comprehensive and detailed reply! That helped explain things for me :) The screenshot especially made it clear.

I knocked up https://github.com/clach04/image_ls to help dump simple metadata for directories of images and CBZ files. Most of mine that I bought from Humble Bundle have the same resolution but I was able to see the problem you described when I created some test images with the differences you mentioned. I'd not realized target width was the width of the most common width in the set!

I have not used the 2panel script before (I've only been using the 2ebook script, which doesn't have the width issue). I did some quick experiments and I also see the target width issue for non-merged (i.e. split only) output as well!

> I am not sure what you meant by "For the border of 4 heuristic." though. Are you referring to the third problem which crops 4 pixels from the sides?

Yes, that what I meant and did not explain clearly.

As a heads up, I just read #327 which implies PR's are not likely to be merged with the current version - so please ignore my request to break up into smaller commits!

PR #412, opened 2022-03-07
file selector add All *.*

StudioEtrange:


darodi:
in which case is this needed?

StudioEtrange:
For file named with wrong extension

Issue #410, opened 2022-03-05
Python Script TypeError: expected string or bytes-like object


HarveyB02:
When running kcc-c2e.exe installed with pip, the following error occurs

`
Traceback (most recent call last):                                                                        File "C:\Users\harve\AppData\Roaming\Python\Python39\Scripts\kcc-c2e-script.py", line 33, in <module>     sys.exit(load_entry_point('KindleComicConverter==5.5.2', 'console_scripts', 'kcc-c2e')())             File "C:\Users\harve\AppData\Roaming\Python\Python39\site-packages\kindlecomicconverter\startup.py", line 46, in startC2E                                                                                         dependencyCheck(2)                                                                                    File "C:\Users\harve\AppData\Roaming\Python\Python39\site-packages\kindlecomicconverter\shared.py", line 121, in dependencyCheck                                                                                  if StrictVersion('1.2.1') > StrictVersion(slugifyVersion):                                            File "C:\Users\harve\AppData\Local\Programs\Python\Python39\lib\distutils\version.py", line 40, in __init__                                                                                                       self.parse(vstring)                                                                                   File "C:\Users\harve\AppData\Local\Programs\Python\Python39\lib\distutils\version.py", line 135, in parse                                                                                                         match = self.version_re.match(vstring)                                                              TypeError: expected string or bytes-like object
`

Any tips?
Windows 11
Python 3.9.9
pip 21.2.4

darodi:
the problem seems related to your slugify version.
try that in the python command line, what is the result?

```python
from slugify import __version__ as slugifyVersion
print(slugifyVersion)
from distutils.version import StrictVersion
print(StrictVersion(slugifyVersion))
```

HarveyB02:
@darodi
```py
>>> from slugify import __version__ as slugifyVersion
>>> print(slugifyVersion)
<module 'slugify.__version__' from 'C:\\Users\\harve\\AppData\\Roaming\\Python\\Python39\\site-packages\\slugify\\__version__.py'>
>>> from distutils.version import StrictVersion
>>> print(StrictVersion(slugifyVersion))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\harve\AppData\Local\Programs\Python\Python39\lib\distutils\version.py", line 40, in __init__
    self.parse(vstring)
  File "C:\Users\harve\AppData\Local\Programs\Python\Python39\lib\distutils\version.py", line 135, in parse
[...]

PR #417, opened 2022-04-20
Update shared.py

ndtoan96:
Fix version of slugify

### DIFF
--- a/kindlecomicconverter/shared.py
+++ b/kindlecomicconverter/shared.py
@@ -117,7 +117,7 @@ def dependencyCheck(level):
         except ImportError:
             missing.append('psutil 5.0.0+')
         try:
-            from slugify import __version__ as slugifyVersion
+            from slugify.__version__ import __version__ as slugifyVersion
             if StrictVersion('1.2.1') > StrictVersion(slugifyVersion):
                 missing.append('python-slugify 1.2.1+')
         except ImportError:


ndtoan96:
Sorry, now I checked again it seems slugify change version structure recently (from v6). It's weird because I install KCC via PyPi and got error at that line, which means PyPi installs the latest version of slugify instead of respecting the dependency mentioned in requirements.txt.

darodi:
@ndtoan96 
Does the new slugify version work fine?


> Sorry, now I checked again it seems slugify change version structure recently (from v6). It's weird because I install KCC via PyPi and got error at that line, which means PyPi installs the latest version of slugify instead of respecting the dependency mentioned in requirements.txt.

I found the problem. [kcc/setup.py](https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84) doesn't include the [requirements.txt](https://github.com/ciromattia/kcc/blob/master/requirements.txt) and there is a mismatch. 
PyPi version was built using [kcc/setup.py](https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84).

https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/requirements.txt#L4

https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84




something like this would be better in [kcc/setup.py](https://github.com/ciromattia/kcc/blob/master/setup.py) 

```python
with open("requirements.txt", "r", "utf-8") as f:
    requires = f.readlines()

setuptools.setup(
...

    install_requires=requires,
```

ndtoan96:
@darodi 
> Does the new slugify version work fine?
Not sure, I got plenty problem with images processing part, the binary works fine though. Will look at it when I have time, maybe due to dependencies just like this one.

darodi:
Duplicate of #411 
Duplicate of #410

Issue #419, opened 2022-04-21
slugify version in setup.py


darodi:
> Sorry, now I checked again it seems slugify change version structure recently (from v6). It's weird because I install KCC via PyPi and got error at that line, which means PyPi installs the latest version of slugify instead of respecting the dependency mentioned in requirements.txt.

I found the problem. [kcc/setup.py](https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84) doesn't include the [requirements.txt](https://github.com/ciromattia/kcc/blob/master/requirements.txt) and there is a mismatch. 
PyPi version was built using [kcc/setup.py](https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84).

https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/requirements.txt#L4

https://github.com/ciromattia/kcc/blob/4ec4c9966c727d6dac44507d34607bd7d2c5ed5c/setup.py#L84

something like this would be better in [kcc/setup.py](https://github.com/ciromattia/kcc/blob/master/setup.py) 

```python
with open("requirements.txt", "r", "utf-8") as f:
    requires = f.readlines()

setuptools.setup(
...

    install_requires=requires,
```

_Originally posted by @darodi in https://github.com/ciromattia/kcc/issues/417#issuecomment-1105453030_

darodi:
Duplicate of #411 
Duplicate of #410

PR #420, opened 2022-04-21
fix: slugify version in setup.py

darodi:
fix: slugify version in setup.py


darodi:
- fix #419

darodi:
Duplicate of #411 
Duplicate of #410

Issue #406, opened 2021-11-30
Work-around to "RuntimeError... 'float' object cannot be interpreted as an integer during execution" for kcc-c2e.py powershell/CLI


aaronshiu:
Hope this is helpful, but I tried implementing the program running kcc-c2e.py for a directory/cbz archive file with powershell on windows. During execution, I kept coming across: 

Processing images...
Traceback (most recent call last):
  File "<Directory>\kcc\kcc-c2e.py", line 32, in <module>
    startC2E()
  File "<Directory>\kcc\kindlecomicconverter\startup.py", line 49, in startC2E
    sys.exit(main(sys.argv[1:]))
  File "<Directory>\kcc\kindlecomicconverter\comic2ebook.py", line 74, in main
    makeBook(source)
  File "<Directory>\kcc\kindlecomicconverter\comic2ebook.py", line 1089, in makeBook
    imgDirectoryProcessing(os.path.join(path, "OEBPS", "Images"))
  File "<Directory>\kcc\kindlecomicconverter\comic2ebook.py", line 526, in imgDirectoryProcessing
    raise RuntimeError("One of workers crashed. Cause: " + workerOutput[0][0], workerOutput[0][1])
RuntimeError: ("One of workers crashed. Cause: 'float' object cannot be interpreted as an integer", '  File "<Directory>\\kcc\\kindlecomicconverter\\comic2ebook.py", line 563, in imgFileProcessing\n    img.autocontrastImage()\n  File "<Directory>\\kcc\\kindlecomicconverter\\image.py", line 258, in autocontrastImage\n    self.image = ImageOps.autocontrast(Image.eval(self.image, lambda a: 255 * (a / 255.) ** gamma))\n  File "C:\\Program Files\\Python310\\lib\\site-packages\\PIL\\Image.py", line 3107, in eval\n    return image.point(args[0])\n  File "C:\\Program Files\\Python310\\lib\\site-packages\\PIL\\Image.py", line 1680, in point\n    return self._new(self.im.point(lut, mode))\n')

For kcc\kindlecomicconverter\image.py, on line 258, 
this was changed from: 255 * (a / 255.) ** gamma
to: int(255 * (a / 255.) ** gamma)

So the line became: self.image = ImageOps.autocontrast(Image.eval(self.image, lambda a: int(255 * (a / 255.) ** gamma)))

This resolved the issue and allowed the program to run. :)

ScrambleTheRamble:
[...]

PR #426, opened 2022-05-28
Fix program to work with new dependency versions

bamless:
The script was crashing in two places due to changes in behaviour of two dependecies (pillow and slugify).
The crash concerning `slugify` was caused by a moved `__version__` variable, which in turn caused the script to fail during dependency checks.
The crash in PIL was due to a lambda that did not round the returned value to int, thus causing PIL to raise an exception (TypeError float object cannot be interpreted as an integer).

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -255,7 +255,7 @@ class ComicPage:
         if gamma == 1.0:
             self.image = ImageOps.autocontrast(self.image)
         else:
-            self.image = ImageOps.autocontrast(Image.eval(self.image, lambda a: 255 * (a / 255.) ** gamma))
+            self.image = ImageOps.autocontrast(Image.eval(self.image, lambda a: int(255 * (a / 255.) ** gamma)))
 
     def quantizeImage(self):
         colors = len(self.palette) // 3
--- a/kindlecomicconverter/shared.py
+++ b/kindlecomicconverter/shared.py
@@ -118,6 +118,8 @@ def dependencyCheck(level):
             missing.append('psutil 5.0.0+')
         try:
             from slugify import __version__ as slugifyVersion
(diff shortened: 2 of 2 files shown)


darodi:
https://github.com/ciromattia/kcc/blob/master/requirements.txt 
should be updated too in your pull request for consistency

### slugify >= 3.0.0 

Duplicate of https://github.com/ciromattia/kcc/pull/411
Duplicate of https://github.com/ciromattia/kcc/issues/410
Duplicate of  #417

### Pillow >=  8.4.0
Duplicate of  #406

bamless:
Actually, I tested the changes with the versions specified in requirements.txt on a conversion using `c2e` and they seem to work fine.
Maybe we could leave the old versions as to not cut off platforms without access to the newest ones?
I would further test the code to ensure the new versions didn't introduce any other regressions, but without unit tests in place its pretty difficult to check every possible execution path. Maybe someone with an in-depth knowledge of the codebase can answer me on that.

bamless:
Also, I'm noticing some inconsistencies in the versions. For example the requirements file specifies a slugify version >=1.2.1 and <3.0.0, but this info does not match the dependency check inside the code (`shared.py::dependencyCheck`, >=1.2.1) nor the dependencies listed in `setup.py`. Any reasons for that?

darodi:
> Also, I'm noticing some inconsistencies in the versions. For example the requirements file specifies a slugify version >=1.2.1 and <3.0.0, but this info does not match the dependency check inside the code (`shared.py::dependencyCheck`, >=1.2.1) nor the dependencies listed in `setup.py`. Any reasons for that?

no reason, just an oversight I guess

https://github.com/ciromattia/kcc/pull/411/commits/4b3588dccd04e244bce12463b01a7cdf046f1f88

darodi:
merged in #457

Issue #307, opened 2019-03-07
Linux binary builds are unavailable


AcidWeb:
Due to migration of build pipeline Linux binary builds are currently unavailable.

Please use PyPI version.

gelatinbomb:
@AcidWeb Hey, sorry for the inconvenience but.....how do i execute this?, i installed with pip3 and i have no idea how to run the kindle comic converter. pls help i'm on linux.

AcidWeb:
If it is installed correctly just execute command: `kcc`

gelatinbomb:
ooh.....that explain it xD
I think it was something like "kindlecomiccreator",  i use kde so...well i have a lot of "k" weird name binaries.

anzetal:
hello @AcidWeb pip3 stil shows 5.4.5. Will the update be pushed out later?

mthsgrc:
~_$ kcc
Traceback (most recent call last):
  File "/home/mthsgrc/.local/bin/kcc", line 6, in <module>
    from kindlecomicconverter.startup import start
  File "/home/mthsgrc/.local/lib/python2.7/site-packages/kindlecomicconverter/startup.py", line 24, in <module>
    from .shared import dependencyCheck
  File "/home/mthsgrc/.local/lib/python2.7/site-packages/kindlecomicconverter/shared.py", line 23, in <module>
    from html.parser import HTMLParser
ImportError: No module named html.parser


Any help?

AcidWeb:
This is Python 3 script.

mthsgrc:
> This is Python 3 script.

Thank you! Newbie Linux User here. 
Changing Python version in use did the trick.

CrownCrafter:
I installed everything... but how do i open kcc?
Typing kcc in the shell gives command 'kcc' not found.
I am on Ubuntu 20.04

masmiguel:
> I installed everything... but how do i open kcc?
> Typing kcc in the shell gives command 'kcc' not found.
> I am on Ubuntu 20.04

Try this:
 I get to install kindle comic converter in ubuntu mate 19 and in linux mint 20:

First of all i put this in the terminal:

sudo apt-get install python3 python3-dev python3-pip python3-pyqt5 libpng-dev libjpeg-dev p7zip-full unrar

sudo pip3 install --upgrade pillow python-slugify psutil scandir raven

After, I dowload the kindlegen from:
https://ia801700.us.archive.org/14/items/kindlegen/kindlegen

And i put in:
[...]

PR #435, opened 2022-09-10
Add Appimage link for Linux binary

CrownCrafter:
x86_64 Appimage support for kcc.
Solves #307.


CrownCrafter:
I will add appimages for other architectures when merged(rpi appimage).

darodi:
Could you provide a pull request that generates an AppImage as part of the project’s build pipeline?
That would be better than just a fixed AppImage that needs to be changed on every source change.

CrownCrafter:
So sorry to sound dumb, but I don't know how to edit the build pipeline?

darodi:
If you can't edit/create the pipeline, can you provide the yml recipe you used to create the AppImage?
So I could include it in an example pipeline for you.
Thanks in advance.

PR #443, opened 2022-11-05
Fix version range used for `slugify` by python "pip" based installation

### LINKED ISSUE #444: pip based installation is not working
While trying to use this interesting tool I stumbled a weird issue when installing using pip (I actually use an outdated mac os version so... not much choice here ;) ). Basically the kcc command line tool would fail starting. See below for details

## Reproduction path

- Install using pip
- Start kcc

Error being raised:
```
  File "/Users/antoine/Library/Python/3.8/bin/kcc", line 8, in <module>
    sys.exit(start())
  File "/Users/antoine/Library/Python/3.8/lib/python/site-packages/kindlecomicconverter/startup.py", line 28, in start
    dependencyCheck(3)
  File "/Users/antoine/Library/Python/3.8/lib/python/site-packages/kindlecomicconverter/shared.py", line 121, in dependencyCheck
    if StrictVersion('1.2.1') > StrictVersion(slugifyVersion):
  File "/usr/local/lib/python3.8/site-packages/setuptools/_distutils/version.py", line 54, in __init__
    self.parse(vstring)
  File "/usr/local/lib/python3.8/site-packages/setuptools/_distutils/version.py", line 155, in parse
    match = self.version_re.match(vstring)
TypeError: expected string or bytes-like object
```

smaarn:
Should fix #444

There were two possibilities here:

1. Fixing the version detection
2. Fixing the version range being used in the setup phase

While the former one could be sufficient to "hide the bug" it wouldn't actually be preventing from using an unsupported version, so I went for the 2nd option here.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -81,7 +81,7 @@ setuptools.setup(
         'PyQt5>=5.6.0',
         'Pillow>=5.2.0',
         'psutil>=5.0.0',
-        'python-slugify>=1.2.1',
+        'python-slugify>=1.2.1,<3.0.0',
         'raven>=6.0.0',
     ],
     classifiers=[],


smaarn:
Actually just read the docs where it says it requires a version lower than 3.0.0... I'm guessing I'll replace this by enforcing the requirements :/

smaarn:
This is actually a duplicate of #411. My bad.

PR #451, opened 2022-12-17
Add default behavior when panels not found

duynl71:
There are some cases that we cannot find the panels and are not returning anything.

So this PR is for adding the whole page and it will be split instead of return nothing

### DIFF
--- a/kindlecomicconverter/comic2panel.py
+++ b/kindlecomicconverter/comic2panel.py
@@ -124,6 +124,9 @@ def splitImage(work):
                     panels.append((panelY1, panelY2, panelY2 - panelY1))
                 yWork += 5
 
+            if len(panels) == 0:
+                panels.append((panelY1, heightImg, heightImg - panelY1))
+
             # Split too big panels
             panelsProcessed = []
             for panel in panels:


darodi:
@duynl71 

same problem as 
- #393  
?

darodi:
@duynl71 

Could you give us an example to check and test the PR?

duynl71:
Hi @darodi here is the sample

If you need to file, please send me your email.
From the code, there is a chance that we cannot find any panel. So this fix is for that case only.

axu2:
What happens if no panels are found?

duynl71:
> What happens if no panels are found?

We are skipping the page

darodi:
@duynl71 

I tried your input file with the last version in master
The conversion works fine without your pull request.

image

Do you have more information?

duynl71:
How about your output? Does it include the page that I posted before?

darodi:
Yes
image

duynl71:
My change is for webtoon mode. Can you try again with webtoon mode?

darodi:
Same result, the conversion is done and I can see the image in the mobi file in sumatra pdf
Is your PR still needed in v5.6.1? Other changes have been merged.
Can you still reproduce the problem with that version?

duynl71:
@darodi hm. Too weird. I am using the v5.5.2 but the same issue as when I debug with the master code.
The settings is:
- kindle PW 1/2 or 3/4 (with default manga mode disabled)
- webtoon mode.
- The rest are default.

If this issue is not happened on your local. You can ignore my PR. I can customize it from my side.

darodi:
Sorry, even with you last parameters and you sample, I'm unable to reproduce.

I'm leaving this open if someone has more information later.

duynl71:
No worry. we can close this PR because of not able to reproduce

PR #459, opened 2023-01-18
Using communicate instead of terminate

catsout:
Fix kindlegen not clear temp files


darodi:
@catsout 
thanks
merged in v5.6.1

PR #461, opened 2023-01-20
Fix pillow backwards compatibility, add mozjpeg-lossless-optimization to setup.py

corylk:
- Fix pillow backwards compatibility - sequence of monkey patch is incorrect
- Add mozjpeg-lossless-optimization to setup.py - supporting installation from github via pip


darodi:
@corylk 
thanks
merged in v5.6.1

Issue #391, opened 2021-04-25
Add Page Fill Option


axu2:
Hey! Longtime mangle user, recent discover of kcc.

Would love a feature like filling screen by cropping when source files are only slightly skinnier.

I implemented it myself for mangle.

https://github.com/FooSoft/mangle/pull/33

It would also be cool to be able to specify where to crop from, though the default is good enough for most cases.

In hindsight, should have called it `fill` not `scaleCrop`

darodi:
Cropping mode has been added in version v5.6.1 is this enough?
If not, please describe what you would like.

axu2:
![image](https://user-images.githubusercontent.com/20757319/214487124-eca11e1b-f5ca-41db-ad5f-18b39b64da9d.png)

Just cropping margins or page numbers is not what I want.

I want to crop the image from the top/bottom or left/right so the remaining image fills the screen.

Here is one example from the PR in my original post:

![image](https://user-images.githubusercontent.com/20757319/214488304-fb59a79e-daec-4101-8325-8bb2efb832aa.png)


In my PR, I replaced the stretch feature with a crop, since I can't imagine anyone wanting to change the aspect ratio.

![image](https://user-images.githubusercontent.com/20757319/214487861-ec1d58fc-a4dc-40fe-bfda-9bf1da943b7a.png)

axu2:
I made a barebones implementation in #471

darodi:
Good idea!

We should maybe rethink the GUI a bit.
There are a lot of checkboxes for everything.
Maybe a simple mode and an advanced mode?

axu2:
Will close this since I already have a pending PR in #495

PR #471, opened 2023-01-26
Crop to fill screen

axu2:
Use ImageOps.fit instead of image.resize

Want to discuss with the team on how to integrate this feature. The barebones implementation is just replacing the stretch to fit feature with a crop to fill screen feature, not sure if team is OK with that. But I don't think people want to change the aspect ratio of their manga.

Solve #391

Example of using cropping feature. Notice the cropped out steam from the Titan and slightly cropped shirt of Eren.

image

For now, I'm running kcc locally to do what I want, thanks!

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -310,8 +310,9 @@ class ComicPage:
         else:
             method = Image.Resampling.LANCZOS
         if self.opt.stretch:
+            self.image = ImageOps.fit(self.image, self.size, method)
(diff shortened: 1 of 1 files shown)


axu2:
Would probably need a new box since multiple options. 

Crop all sides
Crop only top/bottom
Crop only sides

To handle different aspect ratios of single pages and double page spreads.

darodi:
Good idea! 

We should maybe rethink the GUI a bit.
There are a lot of checkboxes for everything.
Maybe a simple mode and an advanced mode?

axu2:
I'll just close this for now until I probably implement it, I no longer use this since I upgraded to the Kindle Scribe.

PR #488, opened 2023-03-02
gui/func: Add a 'delete after conversion' button (closes #458)

### LINKED ISSUE #458: [Feature Request] Delete after conversion
Is it possible having the possibility to delete the file after conversion ? It would really be helpful

Constantin1489:
added 'Delete after conversion' button.
I think the button works but in the full scenario, I'm not sure. 
I just imitated other button objects like `self.croppingBox`.
So please check the code before merging.


kindlecomicconverter/KCC_gui.py
```diff
@@ -294,6 +294,10 @@ def run(self):
             options.maximizestrips = True
         if GUI.disableProcessingBox.isChecked():
             options.noprocessing = True
+        if GUI.deleteBox.checkState() == 1:
+            options.delete = 2
```
Constantin1489, line 298: I'm not sure this part. `options.delete = 2` and so on.

darodi:
@Constantin1489 


Thanks for your PR.

You do not need a tristate in your case and can use 
```
  if GUI.deleteBox.isChecked():
            options.delete= True
```
instead 

I'll try to do the modification when I find some time.

Constantin1489:
Thank you for code reviewing and merging. This was my first experience committing to open-source. I appreciate collaborating with you.

PR #494, opened 2023-03-25
remove transparency from image info

axu2:
kcc will crash if source image has transparency and saving to png. 

Was not a problem in mangle due to Python 2/3 differences. https://github.com/python-pillow/Pillow/issues/3150#issuecomment-401577626

Here's a source image that causes crashes.

Liar_Game-v01_055


darodi:
tested, works fine.
Thanks a lot.

PR #499, opened 2023-04-06
Disable Panel View for kindle scribe

### LINKED ISSUE #498: Disable 4 corner panels?
Is there a way to avoid zooming in on the corners of pages? Accidentally zooming in happens a lot when turning pages. Kindle Scribe is big so I don't need to zoom. 

In Amazon purchased books (example Demon Slayer vol 1), I can disable virtual panel to do this.

darodi:
Disable Panel View for kindle scribe

- fix #498

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1054,6 +1054,9 @@ def checkOptions(options):
     # Splitting MOBI is not optional
     if (options.format == 'MOBI' or options.format == 'KFX') and options.batchsplit != 2:
         options.batchsplit = 1
+    # Disable Panel View for kindle scribe.
+    if options.profile == 'KS':
+        options.panelview = False
     # Older Kindle models don't support Panel View.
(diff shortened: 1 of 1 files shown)


axu2:
I actually suggest reverting this, it seems to autocrop the white space when viewed on device. I'd prefer keeping the whitespace.

darodi:
@axu2 

Even with crop disabled?
We should maybe make "disable panel view" a checkbox.

axu2:
Yea, even though the files aren't cropped in the EPUB and when viewing the final MOBI in the Kindle PC app, they end up cropped when viewed on the actual Kindle. Very strange behavior. Thanks for reverting.

jaroslawjanas:
> @axu2
> 
> Even with crop disabled? We should maybe make "disable panel view" a checkbox.

Any plans for doing this?
Personally, I don't use PV and it gets in the way, so I would very much welcome such an option.

axu2:
@jaroslawjanas Should be fixed in #899, please use the linked test build. It disables panel view

Issue #493, opened 2023-03-24
KindleGen slow for Scribe


axu2:
It takes like 20 minutes for me to run the kindlegen portion even though the epub portion takes like 1 minute. Any way to make it faster? I'm using these settings by the way on a really old PC. I am also running KindleGen on files on my really old HDD, maybe moving the files to the SSD would help?
![image](https://user-images.githubusercontent.com/20757319/227601326-f230340e-4b2f-43ce-9c06-a8fd6f0294be.png)

![image](https://user-images.githubusercontent.com/20757319/227602678-5407f6e8-9c36-4e4d-a93b-949bdf0bcfba.png)

for reference, source files are jpg with 2250 height.

For reference, tried it on my gaming PC and kindlegen ran really fast, so assuming my old CPU is just weak.

darodi:
@axu2 
You could also try to convert to epub and use send to kindle instead.
Here is the link:
web version https://www.amazon.com/sendtokindle
pc version https://www.amazon.com/sendtokindle/pc

axu2:
Oh the limit is now 200 mb now. https://www.reddit.com/r/kindle/comments/z3otof/psa_you_can_now_upload_200_mb_files_to_kindle/

axu2:
Curiously, the older version from archive seems to be significantly faster than the one from Kindle Previewer. Maybe I only encounter the issue since Scribe files are a lot bigger than others.

darodi:
@axu2 

with https://www.amazon.com/Kindle-Previewer/b?ie=UTF8&node=21381691011
using 
` "Kindle Previewer 3.exe" your_book.epub -convert`

is the result faster?

axu2:
Nope, the old version runs in like 2 minutes and the new one runs in like 6. The new one also creates files ~150 mb vs ~100 for the old one. Quite interesting. I don't see a visual difference on my Kindle Scribe.

axu2:
Saw a bug in one of the old-version generated files. A page was in the wrong place, but closing and re-opening the book on the kindle fixed it. Hmmm

axu2:
Overall, I get a much less buggier experience with the files generated by the latest version on my Kindle Scribe.

While it sucks it's a lot slower, less bugs is more important.

axu2:
I just tried generating the mobi at the custom resolution 1440x1920 and it was really fast. Kindlegen transforming source images at the scribe resolution of 1860x2480 down to that really slows it down.

darodi:
[...]

Issue #460, opened 2023-01-19
Kindle Scribe mobi limited to 1920 height


pretenderlu:
When I choose ks for output, the generated mobi does not reach the resolution of 1860*2480, but xxxx*1920.What seems to be a limitation?

darodi:
@pretenderlu 

Do you have an example?
input file and settings used? (GUI settings or command line settings)

pretenderlu:
I have tried many times, whether it is ks or other custom, or choose not to process the picture……maybe you can try any one picture 1860x2480
Sample：[https://drive.google.com/file/d/1SMI5fIb_q0gVkc75bcrg3JgtzOncC2N_/view?usp=share_link](url)

darodi:
@pretenderlu 

I tried your input file
same settings as `Quicker_20230119_163441.png`
if I unzip the generated `001.epub`

in `\001\OEBPS\Images\    `
the file test.jpg is `1860 x 2480  Pixels (4.61 MPixels) (3:4)`

`<img width="1860" height="2480" src="../Images/test.jpg"/>`
in `001\OEBPS\Text\test.xhtml`

Where and how do you see a result of `xxxx1920`?

powersee:
@darodi 
if you turn the epub to mobi, the picture will be change

pretenderlu:
@darodi Epub is OK. The problem is the mobi format file.

darodi:
How did you convert to mobi?

- using kindlegen from https://www.amazon.com/Kindle-Previewer/b?ie=UTF8&node=21381691011 in `Amazon Kindle Previewer 3 Folder\lib\fc\bin`?
- using https://www.amazon.com/sendtokindle with an epub ?

pretenderlu:
@darodi Just use KCC, output epub & mobi as the same time.

pretenderlu:
@darodi I just use KCC, so
 that means I using kindlegen.

The kindlegen is from https://www.amazon.com/Kindle-Previewer/b?ie=UTF8&node=21381691011 in Amazon Kindle Previewer 3 Folder\lib\fc\bin.

nicoladicicco:
I experience the same issue (.epub is 1860*2480, .mobi/azw3 is x*1920) both by using KindleGen and Send to Kindle via browser. :(

darodi:
@pretenderlu @nicoladicicco 

Indeed, there seems to be a limitation with kindlegen.

[...]

PR #500, opened 2023-04-08
dramatically increase Scribe quality/speed

axu2:
Until the issue with Kindlegen #460 is resolved, I suggest lowering the Scribe resolution to the max output of kindlegen

Kindlegen downscaling images makes images soft compared to letting KCC downscale it.

Also, kindlegen runs super slow when downscaling per #493. If KCC has already downscaled it to 1440x1920, it runs instantly. On my PC, it's the difference between 15 second and 6 minute processing.

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -87,7 +87,7 @@ class ProfileData:
         'KV': ("Kindle Paperwhite 3/4/Voyage/Oasis", (1072, 1448), Palette16, 1.8),
         'KPW5': ("Kindle Paperwhite 5/Signature Edition", (1236, 1648), Palette16, 1.8),
         'KO': ("Kindle Oasis 2/3", (1264, 1680), Palette16, 1.8),
-        'KS': ("Kindle Scribe", (1860, 2480), Palette16, 1.8),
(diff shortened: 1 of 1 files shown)


axu2:
@darodi @pretenderlu @nicoladicicco Kindle Scribe support is pretty slow/blurry unless we merge this in. May be worth a hotfix release. Some of the final MOBI files are pretty close to 200 MB even after downscaling.

nicoladicicco:
@axu2 thank you! Indeed, lowering the resolution results in significantly lower processing times and sharper images.
Let's hope that this issue gets eventually fixed on Amazon's side 😅

darodi:
Resolution should be lowered for mobi conversion only.
epub and sendtokindle might work with a higher resolution
see https://github.com/ciromattia/kcc/issues/460#issuecomment-1523816829

Waiting for a result of that test before merging/modifying something.

axu2:
I confirm that send to kindle will compress to 1920 height, if not more, in order to keep each image under 256 kb. So it's pretty compressed.

darodi:
- closed, see #514 
- Resolution should be lowered for mobi conversion only.

PR #502, opened 2023-04-17
revert 499

darodi:
revert 499


axu2:
Appreciated! Too bad the feature didn't work as I expected.

PR #507, opened 2023-05-05
Fix page order and bookmarks by renaming numerically

### LINKED ISSUE #339: Question: ComicRacket metadata page numbering
How does the `Image` attribute for `Page` elements work?

With the config I wrote the bookmark is sometime one or two pages earlier than it should be.

For example, I have

```xml
<Page Image="9" Bookmark="目次"/>
<Page Image="10" Bookmark="其之一：ブルマと孫悟空"/>
```

image 9 is the TOC and image 10 the first chapter, now what's weird is that when I choose to jump to the TOC in my kindle I land two pages earlier, but the bookmark for chapter 1 is correct.

Does this maybe have to do with automatic page splitting and if so is there any way to set the correct image?

### LINKED ISSUE #345: [5.5.2/Win] Large .cbz conversion
After combinging multiple .cbz files with ComicRack, I tried converting a large 1.6GB .cbz file with 5.5.2 windows version of KCC and encountered the following error:

Error during conversion C:\Users\Jan\Desktop\123.cbz:

list index out of range

Traceback:
  File "kindlecomicconverter\KCC_gui.py", line 308, in run
  File "kindlecomicconverter\comic2ebook.py", line 1127, in makeBook
  File "kindlecomicconverter\comic2ebook.py", line 489, in buildEPUB

### LINKED ISSUE #362: Renaming problem
can you add an option for not renaming files it's make some inversion depending the initial name

### LINKED ISSUE #368: List Index Out Of Range
Hi,

i was trying to convert some cbz/cbr files to mobi files (it's comics) but i got an unexpected issue. 

image

If someone can help me it will be a pleasure.

Thanks

### LINKED ISSUE #396: Large files cannot be converted
I found a cbr which is 500 MB ([here](http://booksdl.org/comics0/_NON_ENG_ORIG/%21Incoming/%21%21Tri/%21Dixit/Manga%20FR/G/Grappler%20the%20baki/) if you're interested) and I'm trying to convert it to mobi with KCC. Though the software works great in general, I always get an error when I try to convert this one:
```
Error during conversion! Please consult wiki for more details.
```
I've managed to extract it with online tools (though not to convert it to mobi because all these websites have size limits of 300 MB or lower) so I don't think there is anything wrong with the file.
I haven't seen anything about this in the docs about size limits so I assume this is unwanted behaviour.

### LINKED ISSUE #409: Pages outputted as out of order.
In every file explorer, the pages are in what order they are supposed to be. Windows file explorer and 7 zip have the correct alphabetical order.

I have discovered that this is due to the program putting the character of "#" before a space, which causes stuff like extras to come before the final chapter of a volume.

Who am I kidding this has not been updated in 2 years..

### LINKED ISSUE #418: The bookmark function cannot be used for epub using the new firmware of kobo4.32
The bookmark function cannot be used for epub using the new firmware of kobo4.32

### LINKED ISSUE #556: Error during conversion list index out of range
I was trying to convert the berserk cbz but it keeps showing the same, I tried different sources for the cbz and the one I am using right now has no modifications and is the original. I did the same with Spy x Family from the same source it worked but not for berserk and vagabond. I tried doing it in parts (instead of combining all in one using ComicRack I tried doing in 50 - 70 files and it sill won't work. It works if I convert the chapters individually but not for all-in-one file, the spy x family all in one worked though. Please provide any guidance.
Screenshot 2023-07-06 151521

### LINKED ISSUE #558: Incorrect table of contents if "Rotate and split" is chosen
It seems that if "Rotate and split" is chosen, the table of contents created with ComicTagger is not adjusted correctly. If there are any double pages before the entry, It is off by a few pages (roughly equal to the number of double pages before that page).

For example:
Example file: `https://we.tl/t-RHYm3amWpH`

Settings I used:
image

The first three entries in the ToC is correct (since there aren't any double pages between them). But the fourth one is off by 3 pages (should be 74/171 on Kindle instead of 70/171) because there are 3 double pages before the third and fourth entries. Also, idk why Kindle does not show page 72 (only 70-71-73-74) and the number on the ToC is 68 instead of 70.

axu2:
Fix Page Order Issue:
Fixes #362 
fixes #409

Fix various bookmark issues due to kcc-b logic:
 

 (I suspect larger files made by comicrack are more likely to have bookmarks that cause issues).

fixes #339
fixes #345
fixes #368 
fixes #396 
fixes #418 
fixes #556 
fixes #558 

Don't rename files with slugify and fix the bookmark drift issue. 

Hasn't caused any issues in the files I've tested however, with flat files and subdirectories. 

@cyberbad @4AM-Campfire @VictorGamerLOL @Scotsguy @tao-cumplido @ahihh @noahastor @Jhoelious @crevopunch @augustebaum @origaniels

#46 seems related


axu2:
I've gone ahead and attached a file that can cause issues with this: 
[crop jpg.zip](https://github.com/ciromattia/kcc/files/11860445/crop.jpg.zip)

axu2:
Main issue is with files with a `#` character, which is really low in the ascii sort order.

axu2:
My conclusion: it is impossible to both preserve the original file order and meet EPUB standard for filenames. 

Either give an option to choose one or the other. Or rename files numerically like mangle. (e.g. 0000, 0001, 0002.jpg)

axu2:
Decided to rename pages numerically 0000.jpg 0001.jpg etc

axu2:
This probably has a bad interaction with the `-KCC-A/B/B` at the end of files in the original.

darodi:
@axu2 
is kindlecomicconverter.comic2ebook.sanitizeTreeKobo still needed if all files are renamed with a counter?

axu2:
Oh that's right, we can remove the Kobo one too since they are doing the same thing now.

axu2:
Actually, looks like I need to make a change regarding the bookmarks too with the kcc-b and kcc-c logic. Sorry for missing that.

axu2:
OK, looks like this now: 
image

axu2:
Since this fixes the bookmark drift issue in #339

axu2:
Thank you to @vinhtq115 for extensive testing in #558

PR #370, opened 2020-08-17
lazy hack to fix unrar bug

AlicesReflexion:
This should probably be cleaned up, but I figure I might as well open it now. This fixes #332 for me.

PR #512, opened 2023-05-12
Allow unrar to overwrite files

### LINKED ISSUE #489: Windows cannot convert cbr in version 5.6.1
I can't convert cbr, cbz on Windows 11. KCC version 5.6.1
image

cookie99999:
On systems without the appropriate 7zip functionality, corrupt files will be created in the working directory, causing unrar to fail as its default behavior is to disallow overwriting files with the same names. Fixes #489

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ On Debian based distributions these two commands should install all needed depen
 
 
 ```bash
-$ sudo apt-get install -y python3 python3-dev libpng-dev libjpeg-dev p7zip-full unrar-free libgl1 python3-pyqt5 && \
+$ sudo apt-get install -y python3 python3-dev libpng-dev libjpeg-dev p7zip-full p7zip-rar unrar-free libgl1 python3-pyqt5 && \
     python -m pip install --upgrade pip && \
     python -m pip install --upgrade -r requirements.txt
 ```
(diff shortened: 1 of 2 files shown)


darodi:
@cookie99999 

What linux distribution are you using?
On ubuntu, I get this:

```
# unrar x -y -o+ -x__MACOSX -x.DS_Store -xthumbs.db -xThumbs.db test.rar
unrar: invalid option -- 'y'
Try `unrar --help' or `unrar --usage' for more information.
```

```
# unrar --help
Usage: unrar [OPTION...] ARCHIVE [FILE...] [DESTINATION]
Extract files from rar archives.

  -x, --extract              Extract files from archive (default)
  -t, --list                 List files in archive
  -f, --force                Overwrite files when extracting
      --extract-newer        Only extract newer files from the archive
      --extract-no-paths     Don't create directories while extracting
  -p, --password             Decrypt archive using a password
  -?, --help                 Give this help list
      --usage                Give a short usage message
  -V, --version              Print program version
```

We should maybe revert #370 instead

cookie99999:
I am using Debian 11 (stable). Strangely enough I get the same output as you when trying to use unrar manually, but it works from within the program.
p7zip-rar works fine, so for Debian based systems you don't need unrar as a fallback, however it seems other distros might not have an equivalent package.

darodi:
@cookie99999 

I'm not sure this is needed. The recommended package for linux installation is flatpak and its 7z version manages rar files.
Could you give me an example of problem to test your PR?

darodi:
- related to #515

cookie99999:
I described the problem in my comment on #489 . The unrar fallback is not necessary on Debian based systems, you just have to install p7zip-rar. I was running from source, not using flatpak.

darodi:
thanks, I added your readme changes to #515

Issue #481, opened 2023-02-17
7zip and double page


saisot2006:
Hi, my KCC kept saying no 7zip even though I already installed it. One more thing, KCC mistook my comic pages as double pages though they are all single pages.

Thanx.

darodi:
@saisot2006 

Is the directory of 7z in your `Path` environment variable ?
or is 7z.exe in the directory of KCC?

Please provide an example of your pages to reproduce the double page issue, thanks.

saisot2006:
![07face380fa53de4e2a60313d05b4a94](https://user-images.githubusercontent.com/71863003/219815876-4090a7d3-636a-4b6e-8737-47e794282d2c.jpg)

This is one of my images.

bust4cap:
i think the basic detection is just if the page is wider than it is tall, then it is considered a spread

darodi:
bust4cap gave the answer.
closing this, reopen if needed

PR #513, opened 2023-05-12
add 7z to PATH

axu2:
@cookie99999 @saisot2006 would this fix 7z support on Windows? I notice that when running from source it finds 7z fine but not when running the `.exe` until added to PATH. 

EDIT: When running from source, it uses the 7z in the `other` folder in the repo, not from PATH.

Would prevent issues like #481

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ $ sudo apt-get install qt5dxcb-plugin
 - CB7, 7Z *(With `7z` executable)*
 - PDF *(Only extracting JPG images)*
 
+Add 7z to PATH via `setx path "%path%;C:\Program Files\7-Zip"`
+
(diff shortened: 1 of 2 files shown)


cookie99999:
I'm not at my Windows machine right now, but it was having the same failure with cbz files as well, so PATH does seem to be the issue on Windows.

axu2:
Added message to UI as well. 

Also noticed that when running from source, it uses the pre-packed 7z in the `other` folder in the repo, not from PATH.

cookie99999:
I can confirm that this fixes the issue on Windows. As you noted, the source version will prefer the 7z.exe in the 'other' folder.

darodi (approved):

PR #514, opened 2023-05-13
limit kindle scribe image size to (1440, 1920) when using kindlegen

### LINKED ISSUE #493: KindleGen slow for Scribe

darodi:
limit kindle scribe image size to (1440, 1920) when using kindlegen

- see #500
- fix  #493
- see  #460

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -1153,7 +1153,7 @@ class KCCGUI(KCC_ui.Ui_mainWindow):
         self.versionCheck.start()
         self.tray.show()
 
-        # Cleanup unfisnished conversion
+        # Cleanup unfinished conversion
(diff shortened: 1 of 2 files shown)


axu2:
@darodi Thanks for merging! Though might be nice to enable it for EPUB as well since Send to Kindle EPUBS are also downscaled down to 1920.

darodi:
@axu2 
Even with `Send to Kindle` by email?

axu2:
I use the website since email has a 25 mb limit.

Issue #344, opened 2020-03-01
Metadata editor incorrectly expands ampersands (&)


jamesdanylik:
**_It appears ampersands are inappropriately expanded when used in the Series field of the metadata editor_**.  

I discovered this when trying to set the name field of a `Yotsuba&!` manga.  It allows this input without issue -- but I discovered the output books (and indeed the edited, input book files) have the name `Yotsuba&amp;!`.  It appears the metadata format is XML based; maybe this is related?  

Anyway, this behavior seems to apply anytime an ampersand (&) is used at all, not just in proximity to an exclamation mark (!).

jamesdanylik:
After more testing, this seems unrelated to XML.  When `saveXML` is called with new data present in `self.rawdata`, the incorrect expansion is already there.  If I manually crush it by deleting the string `amp;`, everything works as expected.

Maybe this is from QT someplace?

axu2:
Tried with the qt6 build above still expanding the amp

sillysleepylammie:
![image](https://github.com/ciromattia/kcc/assets/150509997/f40722fe-92c1-48b6-8084-c27fa3516394)
bug is still not fixed in 5.6.5

just-smol:
Following this up to add that apostrophes also expand for me in case that is useful information. Also it's not just in the title but in the epub itself. The content.opf, nav.html, and toc.ncx all have this error. So far it' is only apostrophes and ampersands I have seen, forward slash and other punctuation seems to be unaffected.
![image](https://github.com/ciromattia/kcc/assets/44164901/cc55d383-adc3-4910-b60a-d479eb595450)

Abager01:
Still not fixed in 6.3.1. Both ampersands and apostrophes.

axu2:
For minimal use, KCC 7.1.2 adds a default author GUI option.

For advanced use, maybe https://github.com/comictagger/comictagger is a better option?

If there's issues with KCC reading ComicTagger tags let me know. @jamesdanylik @sillysleepylammie @just-smol @Abager01

axu2:
Fixed in above PR, test buils linked in 

- #923 

@chia591 @jamesdanylik @just-smol @Abager01 @sillysleepylammie

PR #523, opened 2023-05-22
Add native Apple Silicon support by upgrading qt5 to qt6

axu2:
For normal users, prebuilt binary here:
https://drive.google.com/drive/folders/1vlQ4nGlXO1fbNy7cROquEPhZ-Be9Prg-?usp=drive_link

- Fixes https://github.com/darodi/kcc/issues/4

I upgraded to pyside6 which has native Apple Silicon support. (aka a vanilla `pip install pyside6` works).

My process was:

1) Install PySide6 and run
`gen_ui_files.sh`
2) `python kcc.py`
3) Fix the error with a `change all occurences` replacement.
4) Repeat

Porting to PySide6 is actually easier than PyQT6 since PyQT6 dropped pyrcc. PySide6 is also the official library now.

Note that **`KCC_rcc.py`**, **`KCC_ui.py`**, **`KCC_ui_editor.py`**, were autogenerated by pyside6-uic and pyside6-rcc

**`KCC_gui.py`** was manually edited.

image

**This has been the base for testing all my other PRs on Mac, extensively tested.**


axu2:
Ah, yes, instead of .value I should be using enums. https://stackoverflow.com/questions/72161415/qt-checkstate-checked-2-and-qt-checkstate-checked-0

darodi:
The main problem here is that the apple silicon is not supported via github actions.
So it's not included in our automatic build/release process.

pyqt5 is available and works during build.
see: https://github.com/darodi/kcc/issues/4

https://github.com/celogeek/go-comic-converter might be a better solution, it can compile on every platform

but I admit that migrating to pyqt6 (or even another GUI?) will be needed in the end.

darodi:
Note for later, 
if the dependency changes from pyqt5 to pyqt6, also adapt all github actions/builds and docker files

axu2:
Yea, I did this to avoid any hacks during the pip installation process or using Rosetta terminal.

axu2:
Further testing reveals that the file generated by the mac version of kindlegen disables panel view even though I don't want it. This has the side effect of automatically cropping whitespace when viewed on Kindle.

axu2:
@StudioEtrange Do you know how to adapt the `pyrcc5` command to pyqt6?

axu2:
Turns out the problems I had on PyQt6 don't exist on PySide6. Plus, the port was even easier.

axu2:
@darodi Can you test the build files?

axu2:
Got it to build fine on my fork. First time using GitHub actions.

darodi:
Note for later
- #344

could be tested too with qt6

axu2:
Ampersand expansion still occurs in PySide6

axu2:
As an aside, just tried running the PyQT5 version on an Intel Mac on latest macOS Ventura and couldn't get it to run with PyQT5. But the PySide6 version in this PR runs fine.

vinhtq115:
Seems like the checkbox settings are not being saved like previous versions. Drop-down ones are still saved.

axu2:
@vinhtq115 Thank you for pointing that out, should be fixed. It was a remnant of my PyQT6 port

axu2:
Now I understand. I changed that line because it would crash when trying to load the saved settings from qt5. 

But now that qt6 settings are saved I don't need to change it. 

However, I don't want behavior that might mean anyone who ever used the qt5 version can't use the qt6 version. 

I'll look into a way to save the settings separately between qt5/6.

axu2:
Fixed by saving the settings to registry as ints instead of CheckStates

axu2:
Thanks for pointing that out. Should be fixed now.

Issue was due to qt5 saving the checkmark states as ints in the registry. When qt6 expected enums. 

Issue resolved by using ints everywhere, saving enums in the registry seemed to cause issues.

Also changed the QSettings to `(ciromattia, kcc)` so anyone running a previous version of the qt6 build won't have issues in case enums got saved to the registry.

darodi:
When switching from master branch PyQt5 to pyqt6 branch, all previous checkboxes values are lost.
That might be a problem when users will upgrade. 

it will have to be reported in the release note at least.

axu2:
That's a good point, I'll save the settings in the same location as qt5, since I'm still using ints.

axu2:
MozJpeg errors were due to a bugged python version. Removed.

darodi:
beta release here
https://github.com/darodi/kcc/releases/tag/v5.7.0b1

Windows is ok
I'm trying appImage now and docker
No apple to test dmg. Can you try @axu2  ?

darodi:
appImage is broken

```
(user@linux kcc-appImage)$ ./kindleComicConverter-latest-x86_64.AppImage 
  File "kcc.py", line 74, in <module>
  File "kindlecomicconverter/startup.py", line 29, in start
  File "PyInstaller/loader/pyimod02_importers.py", line 385, in exec_module
  File "kindlecomicconverter/KCC_gui.py", line 30, in <module>
ImportError: libQt6Widgets.so.6: cannot open shared object file: No such file or directory
[13995] Failed to execute script 'kcc' due to unhandled exception!
```

axu2:
Works fine on Apple. 
image

Surprisingly, the UI is now a lot smoother, the pyqt5 was a little sluggish when running through Rosetta.

darodi:
Docker image is fine

```

user@wsl:/mnt/c/Users/rodi_/workspace/mangas/testmanga$ docker run --rm -v "$(pwd):/app" ghcr.io/darodi/kcc:5.7.0b1 -p KPW5 ./test
comic2ebook v5.7.0b1 - Written by Ciro Mattia Gonano and Pawel Jastrzebski.
Working on ./test...
Preparing source images...
Checking images...
Processing images...
Creating EPUB file...
Creating MOBI files...
```

and mobi output can be read

darodi:
so just the appImage to fix. I'll have a look.

darodi:
docker is broken on amrv7 too

user@amrv7 docker # docker run --rm -v "$(pwd):/app" ghcr.io/darodi/kcc:5.7.0b1 -p KPW5 ./test
ERROR: psutil 5.0.0+, python-slugify 1.2.1+, Pillow 5.2.0+ is not installed!

axu2:
I know nothing about building on Linux so can't help. 

Btw does GitHub show download counts of all the different platforms?

axu2:
I did bump the psutil version due to security but maybe I bumped it too high?

darodi:
I'll have a look and update the Docker image next week

axu2:
@darodi 

On posix isn't it supposed to be in /usr/bin/local or something similar?

darodi:
same problem with version 5.6.2 in fact
`/usr/local/bin` is a read only file system on the steam deck
I'll try with another machine

axu2:
@darodi 

kcc.py has the code that adds 7z and KP3 to path on windows and mac. Only windows adds the kcc directory to path. Maybe you can alter it for Linux.

PR #532, opened 2023-06-03
Override book's title if `Title` is set in XML

vinhtq115:
In ComicRack XML Schema, there is a `Title` field that is not being used by KCC. In this PR, I changed so that if this field is set, it will automatically be used.

Also, `pyinstaller` is added to `requirements.txt` for [building on M1](https://github.com/darodi/kcc/issues/4#issuecomment-1364553511) to avoid `pyinstaller: command not found`.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -695,7 +695,9 @@ def getComicInfo(path, originalpath):
             os.remove(xmlPath)
             return
         options.authors = []
-        if defaultTitle:
+        if xml.data['Title']:
+            options.title = hescape(xml.data['Title'])
+        elif defaultTitle:
             if xml.data['Series']:
                 options.title = hescape(xml.data['Series'])
             if xml.data['Volume']:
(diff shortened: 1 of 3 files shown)


darodi:
@vinhtq115 
Thanks for your PR.

imho, we shouldn't add `pyinstaller` to `requirements.txt`
It is only needed for building an executable.
It is not needed when using the project from source
and in docker images, for example, it would add unneeded dependencies.

vinhtq115:
I have removed `pyinstaller` from `requirements.txt`. Since the [building instruction for M1 ](https://github.com/darodi/kcc/issues/4#issuecomment-1364553511) did not mention about `pyinstaller` and `appdmg`, I suggest adding a Wiki page for users who want to build the binary themselves.

darodi:
@vinhtq115 

https://github.com/ciromattia/kcc/wiki/How-to-build-M1

vinhtq115:
@darodi I have updated the wiki page to be more specific.

darodi:
Thanks for the PR

PR #546, opened 2023-06-18
skip single pixel images in PDF

### LINKED ISSUE #540: [Bug] Every Other Page are Gray Pages(PDF Issue)
Hi,

I'm trying to convert a manga from PDF to EPub and every time I convert, every other page in the epub file contains a page filled with gray. I'm using a Kobo Clara HD with Manga Mode, Stretch/Upscale, and Cropping Mode with the power at 1.00x enabled. Is this a bug or is there an option that I can enable to get rid of this issue?

Eg. Every page has the actual manga content like so: 
image

But then a gray page like this appears:
image

And this pattern repeats for the whole ePub file.

axu2:
Fixes #540

### DIFF
--- a/kindlecomicconverter/pdfjpgextract.py
+++ b/kindlecomicconverter/pdfjpgextract.py
@@ -61,9 +61,11 @@ class PdfJpgExtract:
             istart += startfix
             iend += endfix
             jpg = pdf[istart:iend]
-            jpgfile = open(self.path + "/jpg%d.jpg" % njpg, "wb")
-            jpgfile.write(jpg)
-            jpgfile.close()
-            njpg += 1
+            # hopefully skip single pixels
+            if (iend-istart - 160) > 4:
+                jpgfile = open(self.path + "/jpg%d.jpg" % njpg, "wb")
+                jpgfile.write(jpg)
+                jpgfile.close()
+                njpg += 1
             i = iend
         return self.path, njpg
--- a/kindlecomicconverter/pdfjpgextract.py
+++ b/kindlecomicconverter/pdfjpgextract.py
@@ -25,6 +25,11 @@ import os
 from random import choice
 from string import ascii_uppercase, digits
 
+# skip stray images a few pixels in size in some PDFs
+# typical images are many thousands in length
+# https://github.com/ciromattia/kcc/pull/546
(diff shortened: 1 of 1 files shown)


Derzz:
Ok so good news is that it fixed the majority of the pages. However, there are still some pages in my manga that are showing these gray pixels again. Here's the pages that are having this issue again. Nonetheless, thank you for your work so far!
[よつばと! 第01巻_removed.pdf](https://github.com/ciromattia/kcc/files/11804455/01._removed.pdf)

axu2:
Can you add this line to the code:

`print(iend-istart)` 

and adjust the thresholds as appropriate based on what you see?

Derzz:
Ok, I got it now, I changed the boundaries to 287 since some image lengths were being to that number. Nonetheless, skipping images by these sizes should be fine since these are small values compared to the manga's pages whose full length should be in the thousands. Thank you so much for your help! :)

axu2:
Feel free to make a proper PR when you are done testing. :)

axu2:
@Derzz Ended up fixing it up, ready for review.

kindlecomicconverter/pdfjpgextract.py
```diff
@@ -25,6 +25,11 @@
 from random import choice
 from string import ascii_uppercase, digits
 
+# skip stray images a few pixels in size in some PDFs
+# typical images are many thousands in length
+# https://github.com/ciromattia/kcc/pull/546
+STRAY_IMAGE_LENGTH_THRESHOLD = 300
```
darodi, line 31: > Ok, I got it now, I changed the boundaries to 287 since some image lengths were being to that number. 

_Originally posted by @Derzz in https://github.com/ciromattia/kcc/issues/546#issuecomment-1599315223_

@axu2 
shouldn't it be lower than 300?

darodi (changes requested):

kindlecomicconverter/pdfjpgextract.py
axu2: I used 300 since I saw 159 160 287 in testing and thought it was a reasonable threshold. Real images are many thousands in length so definitely won't overlap.

darodi (commented):
testing this and then merging

darodi (approved):

PR #491, opened 2023-03-11
Updates GUI text for new Homebrew version

thatrobotdev:
The instructions in the GUI for downloading the kindle-comic-creator cask utilize an old brew command, so this PR updates the GUI text with the command that Homebrew uses now. It also renames references to "Brew" to "Homebrew".

Screenshot of the GUI with the old command

---

This is what current homebrew tells you if you use the old command that the GUI suggests:
```shell
jameskerrane@Jamess-MBP ~> brew cask install kindle-comic-creator
Error: `brew cask` is no longer a `brew` command. Use `brew <command> --cask` instead.
```

This is a successful installation with the command that I changed the GUI instructions to:

```shell
jameskerrane@Jamess-MBP ~ [1]> brew install --cask kindle-comic-creator
==> Downloading https://kc2.s3.amazonaws.com/KindleComicCreatorInstall.dmg
Already downloaded: /Users/jameskerrane/Library/Caches/Homebrew/downloads/29bff483a8f694859b20eacf3cd654d0565c07fc31165f9ba8408c3f88edb26a--KindleComicCreatorInstall.dmg
Warning: No checksum defined for cask 'kindle-comic-creator', skipping verification.
==> Installing Cask kindle-comic-creator
==> Running installer for kindle-comic-creator; your password may be necessary.
Package installers may write to any location; options such as `--appdir` are ignored.
Password:
installer: Package name is Kindle Comic Creator
installer: Installing at base path /
installer: The install was successful.
🍺  kindle-comic-creator was successfully installed!
```

PR #555, opened 2023-07-05
refactor display when kindlegen is missing

darodi:
refactor display when kindlegen is missing
display_kindlegen_missing

- fixing links
- same display called in 2 contexts

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,11 @@ $ sudo apt-get install -y python3 python3-dev libpng-dev libjpeg-dev p7zip-full 
 $ sudo apt-get install qt5dxcb-plugin
 ```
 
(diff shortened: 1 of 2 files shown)


darodi:
- fix #491

axu2:
I tried running this from source on Mac.

Just brew installing the cask is not enough. You also need to copy kindle gen to /usr/bin/local, like the wiki says.

I recommend we delete the OS specific code and just link directly to the wiki, similar to how we do for 7z.

https://github.com/ciromattia/kcc/wiki/Installation#kindlegen

axu2:
In retrospect, it was probably because only Kindle Previewer was added to PATH in the KCC code and not comic creator...

axu2:
Superseded by #566 due to merge conflicts

PR #561, opened 2023-07-20
Add de-dupe cover option for landscape alignment

### LINKED ISSUE #433: Empty cover page
Is it possible for a option to add a empty cover page so that when reading on landscape the pages will match? I'm not sure how to explain but for example if there's 1 spread page that's splitted it won't show when on landscape, it will show 1/2 of the spread page + 1 random page.
I hope this makes sense.

### LINKED ISSUE #557: First page duplicated
When a convert my .cbz manga the first page always gets duplicated. When I open the book on kindle it enters automatically on the second page so it doesn't affect me that much but I would like to don't get this duplicated pages only for preciousness if possible. The project is awesome! Congrats!

### LINKED ISSUE #640: [Feature Request] Add option for the insertion of a single blank page to correct alignment of two-page spreads
I don't know if it's something inherent to Kobo devices or to the conversion process, but the first page of any file created with KCC, at least when viewed on a Kobo Libra 2, takes up two pages. This is problematic when the images in the archive were already arranged so that two-page spreads would properly fall on two pages. This causes the images to fall out of alignment and all two-page spreads to be split between one page and the next. It's possible to remedy this by extracting the archive and duplicating two pages near the beginning so the alignment is fixed, but that takes significantly more time (especially when converting numerous archives) than **a checkbox option that would insert a single blank page and fix the alignment during the conversion process,** no extraction necessary.

If, as I think, this is a quirk of Kobo devices, an option such as this could save many people a lot of time.

resendeBruno:
- fix #433 
- fix #557 
- fix #640 

~~--prefercoverfile use cover.jpeg (or cover.png) as the cover instead of the first item on the folder~~

--nocoveraspage deletes the image selected to be a cover so it doesn't become a page inside the book. Useful when the epub reader shows the cover as the first page 

~~--author allows you to name the author yourself~~


resendeBruno:
--nocoveraspage probably fixes this https://github.com/ciromattia/kcc/issues/557

axu2:
Related #433

axu2:
Out of curiosity we're you testing on a Kobo? @resendeBruno

axu2:
Could prefer cover just be the default behavior instead of an option?

Can remove cover be a gui option too? We've upgraded to qt6.

PR #564, opened 2023-07-29
fix mac cover upload when frozen by removing plist

axu2:
Fixes #482

Fixes #436 together with #565

To clarify frozen is referring to a pyinstaller built binary

EDIT: Various issues that turned out to be unrelated were deleted.

PR #565, opened 2023-07-30
fix 7z on Mac when frozen by removing old binaries

### LINKED ISSUE #436: MacOS Monterey v12.6: “Kindle Comic Converter” cannot be opened because the developer cannot be verified.
When I try to open the KCC App on my MacBook I get the following message:
Screenshot 2022-09-16 at 17 15 29

“Kindle Comic Converter” cannot be opened because the developer cannot be verified.
macOS cannot verify that this app is free from malware.

### LINKED ISSUE #528: Error During Conversion For All CBZ Files | Mac OS
Adding CBZ file > EPUB > Convert: "Error during conversion!" but no specific reason as to why

Toggled on for Kindle PW 11 + Manga Mode, All Files are under 500 MB. First time using the converter so any fixes?

axu2:
7z on Apple Silicon is stored in a different location. 

Also removed the binaries so users must install their own 7z from homebrew.

The old versions don't work on either Intel or Apple Silicon on latest macOS. If we leave them there KCC will detect them and not ask the user to install a more updated version of 7z. So any CBZ conversions will fail without a good error.

fixes #528 
Fixes #436 together with #564

@Beebeedeedop @chenobies @sparkirat @maxthiel


axu2:
> removing the included binaries will complicate things for those using them with older mac os architectures.

I don't think it's a problem to ask users to install their own 7z, especially if it will fail without a good error message.

axu2:
Superseded by https://github.com/ciromattia/kcc/pull/566 due to merge conflicts

Issue #392, opened 2021-05-15
How to enable landscape mode when reading manga on kindle?


Kolakona:
As per title. I am converting the jpeg files to mobi using KCC but I am unable to rotate it to landscape mode. 
when use old version like v2.7 ,it works.But in v5.5.2,the converted mobi file can't be read in landscape mode.
Thanks!

darodi:
> the converted mobi file can't be read in landscape mode

What is expected and what is the current behaviour?

axu2:
Epubs from humble bundle have landscape orientation when using Send To Kindle. So it means I can read two pages side by side in landscape mode. There's probably something in the epub metadata that enables this. I don't have epub experience myself.

darodi:
- fixed with #567

PR #567, opened 2023-07-31
Enable double spread on kindle

swaggy-p-jp:
Following this issue https://github.com/ciromattia/kcc/issues/392, wrote the same metadata used for double spreads on the kobo for the kindle as well. Worked as intended. Below are screenshots on the KW5. 

screenshot_2023_07_31T18_38_04-0500
screenshot_2023_07_31T18_38_09-0500
screenshot_2023_07_31T18_38_21-0500
screenshot_2023_07_31T18_38_29-0500

Part of the specification for epubs mentioned here https://github.com/kobolabs/epub-spec#synthetic-spreads, so there shouldn't be an issue for any other ereader either if someone wants to make those changes.


darodi:
@swaggy-p-jp 

thanks for the PR!

axu2:
Wow, thanks for restoring this feature!

PR #574, opened 2023-08-08
prioritize KC2 over KP3 kindlegen version

### LINKED ISSUE #573: KCC Folder
Hello. 
This installation wiki says:

Move kindlegen.exe from Kindle Previewer into the same folder as KCC.

But when I download the installer, I don't see the folder. It's just the KCC app that opens. Where is the KCC folder?

For reference I am on windows and ran this installer. 
https://github.com/ciromattia/kcc/releases/download/v5.6.2/kcc_5.6.2.exe

axu2:
Since KC2 will ask you to install KP3, I'm sure some user will do that.

This is a potential problem since KP3-kindlegen on windows sometimes hangs, so may as well prioritize the older version from KC2. 

KCC v5.6.3 does not search for KP3 or KC2 on Windows. Will need a new release since we merged in #571  afterward. I think installing KC2 is way simpler than moving `kindlegen.exe` from web archive around.

image

links here: https://github.com/ciromattia/kcc/wiki/Installation#kindlegen

Fixes #573

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,6 @@ $ sudo apt-get install qt5dxcb-plugin
 - KindleGen ~~[(deprecated link)](http://www.amazon.com/gp/feature.html?ie=UTF8&docId=1000765211)~~ v2.9+  (For MOBI generation) 
   - should be placed in a directory reachable by your _PATH_ or in _KCC_ directory
   - `KindleGen` can be found in [Kindle Previewer](https://www.amazon.com/Kindle-Previewer/b?ie=UTF8&node=21381691011)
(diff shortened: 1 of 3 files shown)


darodi:
@axu2 

I'm using kindlegen from Amazon Kindle Previewer 3 version 3.71.1 for months without problem.
The buggy version seems recent.
How do you get the bug?

I removed the link to kindlegen from kcc issues in the wiki.
Amazon's licence doesn't allow distribution of kindlegen. It can only be downloaded from Amazon.

axu2:
@darodi I just convert a bunch of comics in a row and it will start failing with the version currently included in Previewer. I will post the error message when I reproduce it.

axu2:
@darodi There's a small chance you never noticed because I'm using the Kindle Scribe profile which is probably bigger than the device you are using. The settings I used are in the screenshot.

PR #575, opened 2023-08-08
remove old windows 7z binaries and Inno Setup Script ISS

axu2:
These were only used when running from source, no need to keep these old binaries around since we now add 7z to path on windows already.

And that Inno Setup Script is ancient and definitely not used anymore, don't want to maintain it since it references the 7z


darodi:
indeed, time for "spring-cleaning" :)

PR #581, opened 2023-08-17
fix files with ' or " by using arg lists instead of strings

axu2:
Now things are a lot more readable and secure. No shell early quote ends.

Also replaced all uses of `Popen` and `communicate`

May also fix the kindlegen from kp3 error, haven't replicated.


darodi:
@axu2 
Master branch

image

your PR:

image

This doesn't seem to work

darodi (changes requested):
not working, see 
https://github.com/ciromattia/kcc/pull/581#issuecomment-1781753498

axu2:
Whoops, didn't notice since that error got bypassed on mac due to `unar`

I'll also improve the error logging.

image

darodi (approved):

PR #589, opened 2023-09-08
fix page order in pdf and more by using natural sort instead of python sort

### LINKED ISSUE #587: Conversion Issue: Pages out of order
Hi I'm new to KCC and I tried to use it to convert a volume of One Piece that I wanted to upload to the kindle, however, despite the fact that I created a special folder for the volume and put inside the chapter folders, when I go to convert it I get a mobi file with all the messy pages.
(I tried to convert the folder with both a directory and the .zip format of it but it always gives me the same problem)
I hope you can help me!  (I use KCC 5.6.3)
Screenshot 2023-09-07 alle 20 27 50
Screenshot 2023-09-07 alle 20 27 55
Screenshot 2023-09-07 alle 20 28 04

### LINKED ISSUE #595: Images showing up in wrong order
Title. Images from a zip and pdf are either skipped or put in the wrong order. Small chapters work fine but volumes of manga don't work right. This would be fine if I could merge .mobi files insome way but i can't find anything online

axu2:
Fixes #587
Fixes #595

So that's why we did number padding in the old sort method. 

Related: #507 

Prebuilt binaries for download for normal user testing:

Windows EXE: https://github.com/axu2/kcc/actions/runs/6123144358
Mac DMG: https://github.com/axu2/kcc/actions/runs/6123143150
Linux: https://github.com/axu2/kcc/actions/runs/6793159103

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -31,6 +31,7 @@ from tempfile import mkdtemp, gettempdir, TemporaryFile
 from shutil import move, copytree, rmtree, copyfile
 from multiprocessing import Pool
 from uuid import uuid4
+from natsort import natsorted
 from slugify import slugify as slugify_ext
 from PIL import Image
 from subprocess import STDOUT, PIPE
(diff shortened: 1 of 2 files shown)


darodi:
Thanks @axu2 
I'll adapt dockers and such before merging this.

axu2:
@darodi Could we get a hotfix v5.6.4 release with this included (and maybe some of the other easier pending PRs)? I see a little too many complaints about this related to pdf especially.

Looks like I broke pdf image extraction order too since our pdf extraction method names files without padding (0.jpg, 1.jpg... 10.jpg, etc.)

elliopitas:
had problems with page number and this fixed it

axu2:
@darodi could we get a mini-release with this fix included?

darodi:
merging this.
I'll test docker build later and create a PR at that moment if this is needed

PR #590, opened 2023-09-08
replace exit() with sys.exit()

### LINKED ISSUE #588: Exit error
I ran into this error message after I tried opening kcc version 5.6.3.

I use windows 10 and the issue is not my antivirus-software, because I already tried downloading and opening it with the software disabled.

I also tried the non-docker version but it still didn't work.

image

axu2:
Fixes #588 @DemonEye888

Temporary EXE here: https://github.com/axu2/kcc/actions/runs/6125895319

Please test to confirm it works.

### DIFF
--- a/kcc-c2e.py
+++ b/kcc-c2e.py
@@ -22,7 +22,7 @@ import sys
 
 if sys.version_info < (3, 8, 0):
     print('ERROR: This is a Python 3.8+ script!')
-    exit(1)
+    sys.exit(1)
 
 from multiprocessing import freeze_support, set_start_method
 from kindlecomicconverter.startup import startC2E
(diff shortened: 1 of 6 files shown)


darodi:
isn't `import sys` needed?

axu2:
@darodi added and force pushed. It's already imported in most files though.

axu2:
Mind finalizing v5.6.3 as is so we don't get questions in the future from people who initially downloaded v5.6.2 since it shows up first like in #591 v5.6.3 has proved pretty stable so far.

image

However, don't include this PR in it. I finally managed to replicate this.

Fixing this issue now makes the app exit silently instead of throwing an error 😓 

I placed KCC in the KP3 directory.

DemonEye888:
I can only test it at the end of next week because I am not at home right now.
But I will when I get the chance.
Thanks again for the help!

darodi:
> @darodi Mind finalizing v5.6.3 

@axu2 
done

PR #604, opened 2023-10-07
fix list index out of range for comicrack

### LINKED ISSUE #582: List index out of range
Hey guys,

Got an issue with 5.6.3 that i thought has been resolved. My issue links with #556  where i get the list 'index out of range'.

I saw that this issue was linked to #507 which is now closed. I do use ComicRack to combine the files into one large book. 

Screenshot-2023-08-14-165430

Test file: https://www.dropbox.com/scl/fi/rvapeciwd71s86dtdf251/Gantz-1-5.cbz?rlkey=9o1peile4b7gjbbbql08l6m4j&dl=0

### LINKED ISSUE #586: list index out of range
had this issue while using kcc 5.6.2
image
image

Ger4ldK:
Fixes #582
Fixes #586

The purpose of this PR is to fix the issues encountered in #582, from what I was able to find based on the OP's file and my own files when using the tool, it seems that the source files we have contain 2-in-1 spread pages, which affects how the index counts up when working with the Comicrack XML Pages/Bookmarks tag. I have added a checkbox under Manga mode(as it seems quite common in manga series to have 2-page spreads) to toggle this option and adjust the math in the loop when converting files. If this checkbox is checked, the delta is adjusted when the Spreat splitter is either Unchecked or Indeterminate (state 0 or 1), and in the case where 'Rotate' is chosen (state 2), the pageid is decremented by 1 each time it encounters a spread to move the bookmark to the correct location. In all 3 states, I did not encounter the error again and my bookmarks were also in the correct positions

I am not sure if this is something that everyone is going to be using, but I will make this PR as a reference for anyone who is encountering the issue faced in #582, I've also tried to maintain the previous behavior as best I can. I am quite new to open source work so feel free to provide any feedback!


axu2:
I'm personally against making users need to check a box to avoid ComicRack related errors. Normal users shouldn't need to check if their source files are merged are not.

Is it possible to do something like:

```python
try:
    original_method()
except IndexError as e:
    your_method()
```

@vinhtq115 Can you confirm this doesn't cause a regression with #558? I deleted the source file.

Ger4ldK:
Yep, I also don't think its the best idea to have this extra checkbox, will try the except IndexError method while I'm converting some more of my library, thanks for the work you guys put into the project!

axu2:
Please add the following text to the first post so GitHub tags it properly:

> Fixes #582 
> Fixes #586 

And change the title to "fix list index out of range for comicrack"

Ger4ldK:
updated the title and first post, but in terms of the fixing, i think i found that the IndexError isn't raised on some other books I was converting - these books didn't have a bookmark in the last few pages, which i think is what causes the calculation to go OOR. The error is no longer showing, but I'm getting inconsistent bookmarks again for books that do not have a bookmark near the final pages

axu2:
Oh hmmm, so the error isn't guaranteed, that's good information.

Maybe go back to original implementation? Can you determine `options.alreadymerged` based on the filenames by looking for `kcc-a`?

Ger4ldK:
Yep, currently working and testing on determining the conditions for the different method by comparing the original file count against the resulting pages after pages are split - a 'true' page count, will update soon

axu2:
Thank you for the cleanup, much easier to review. 

Am I missing something or does this code not preserve the previous behavior? I thought this was an issue related to large files joined by comic rack but these changes appear to apply to all files with comic rack bookmarks.

vinhtq115:
> @vinhtq115 can you provide a test file as well? I want to make sure we don't break your use case. I unfortunately already deleted the one you provided in #558 I believe the file was a complete english official volume with spreads joined. The current code placed the bookmarks correctly for that file.

[Here](https://we.tl/t-yukzZBaF3w) is the test file.

Ger4ldK:
> Thank you for the cleanup, much easier to review.
> 
> Am I missing something or does this code not preserve the previous behavior? I thought this was an issue related to large files joined by comic rack but these changes appear to apply to all files with comic rack bookmarks. Unfortunately I have no comicrack files to test myself.

Oh, I might have misunderstood the splitcheck code then, but my assumption is that when a split is needed, that likely meant that a spread was joined in the source file, which is why i set the `options.hasspreadpages` flag to True. It's going to run on all CR files that detected a split, but if they don't, it should behave as it did previously. I could duplicate the code and maintain the old version too for easier readability. Let me know if I should do that.

axu2:
@Ger4ldK Be aware that the provided test file is NSFW. Be careful when you open it.

Ger4ldK:
> @Ger4ldK Be aware that the provided test file is NSFW. Be careful when you open it, though please report how it interacts with your changes.

thanks for the heads up 😆 my own test files are NSFW too, so no worries, will report back with my results

Ger4ldK:
Tested with @vinhtq115's file, and it seems to be following the pages defined in comicinfo.xml, which means it's working, but the bookmarks defined by them are not the same as the ones in the ToC, which is what I did for my own files, so I'm not sure if this is working for them. Might have to redesign this a bit. I don't have ComicRack and just wrote the XML based on the example in the docs, which might make a huge difference in how this all behaves now

@vinhtq115 was the page numbers adjusted in the test file due to the issues in #558?

vinhtq115:
> Tested with @vinhtq115's file, and it seems to be following the pages defined in comicinfo.xml, which means it's working, but the bookmarks defined by them are not the same as the ones in the ToC, which is what I did for my own files, so I'm not sure if this is working for them. Might have to redesign this a bit. I don't have ComicRack and just wrote the XML based on the example in the docs, which might make a huge difference in how this all behaves now
> 
> @vinhtq115 was the page numbers adjusted in the test file due to the issues in #558?

I bookmarked it as I had done in #558 (no differences).

Ger4ldK:
> > Tested with @vinhtq115's file, and it seems to be following the pages defined in comicinfo.xml, which means it's working, but the bookmarks defined by them are not the same as the ones in the ToC, which is what I did for my own files, so I'm not sure if this is working for them. Might have to redesign this a bit. I don't have ComicRack and just wrote the XML based on the example in the docs, which might make a huge difference in how this all behaves now
> > @vinhtq115 was the page numbers adjusted in the test file due to the issues in #558?
> 
> I bookmarked it as I had done in #558 (no differences).

hmm, might rework my design slightly then, will work with your test file since it helps me understand. Thanks for the info!

Ger4ldK:
Update, managed to get it working with the testing file, and it does not have issues with mine as well

Summarizing the fix as best I can here, this should fix the error of the index going OOR, and additionally it also fixed issues with bookmark drift that I encountered due to my manual editing of my ComicInfo.xml file. Based on my testing with both files and all 3 possible states, the bookmarks all appear correctly for me, I'm no longer encountering issues with the index going OOR when the bookmark is near the end, where `((pageid + global_diff) > len(filelist))` as well

axu2:
@Ger4ldK Please confirm the results with the LARGE comicrack joined file from #582

Ger4ldK:
Sorry for the delay, tested with the file in #582 and it seems to be fine now, no index error. Will generate the binaries and let the OP of the issue test as well to confirm

axu2:
You just link the individual runs.

axu2 (changes requested):
Some small nitpicks.

kindlecomicconverter/comic2ebook.py
```diff
@@ -525,16 +526,26 @@ def buildEPUB(path, chapternames, tomenumber):
         elif options.splitter == 2:
             diff_delta = 2
 
+        if options.spreadadjust:
+            if (options.splitter == 0 or options.splitter == 2):
```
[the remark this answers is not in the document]
Ger4ldK: updated to `if options.spreadadjust and options.splitter != 1:`

kindlecomicconverter/comic2ebook.py
```diff
@@ -581,10 +592,13 @@ def imgFileProcessingTick(output):
         workerOutput.append(output)
         workerPool.terminate()
     else:
-        for page in output:
+        for page in output[0]:
```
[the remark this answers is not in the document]
Ger4ldK: ooh, thanks for this, let me rewrite it

Ger4ldK:
> You just link the individual runs.
> 
> Make sure you tick Stretch/Upscale in the test file so it gets split after 400 mb. Replicate his exact settings to trigger 400 mb split.

was wondering why i wasn't getting a 400mb split even with the same settings, realized i had to select a Kindle device and got this now 
image
Any idea what this is caused by? The line at 218 and the error message isn't giving me much info

axu2:
The error is more easy to see if you revert your changes. 

I feel like there are multiple issues here. 

1) @vinhtq115 where did you get the bookmarks for your test file? Did you manually add? Your test file already worked prior to this PR. 

2) the large test file had bookmarks generated from comicrack directly from #582. I thought this pr fixed it. Fixing that test file is very important. 

3) @Ger4ldK you manually added bookmarks? This is tricky if you make them different than comicrack. 

Honestly we should probably just skip the bookmark related code if a 400 mb split is detected.

vinhtq115:
> 1. @vinhtq115 where did you get the bookmarks for your test file? Did you manually add? Your test file already worked prior to this PR.

I manually created it using ComicTagger.

kindlecomicconverter/comic2ebook.py
```diff
@@ -513,6 +513,7 @@ def buildEPUB(path, chapternames, tomenumber):
                                                    tomenumber), options.uuid))
     # Overwrite chapternames if tree is flat and ComicInfo.xml has bookmarks
```
axu2, line 514: Maybe this entire code section should be skipped if a 400 mb split is detected?

kindlecomicconverter/comic2ebook.py
```diff
@@ -525,16 +526,25 @@ def buildEPUB(path, chapternames, tomenumber):
...
+                if x < filelen:
+                    if '-kcc-b' in filelist[x][1]:
+                        pageid += diff_delta
+                        global_diff += diff_delta
+                    if options.spreadadjust and options.splitter == 1 and '-kcc-a' in filelist[x][1]:
+                        pageid -= 1
+                        global_diff -= 1
+
+            if pageid >= filelen:
+                pageid = filelen - 1
```
axu2, line 547: pageid = max(pageid, filelen -1)

kindlecomicconverter/comic2ebook.py
```diff
@@ -513,6 +513,7 @@ def buildEPUB(path, chapternames, tomenumber):
                                                    tomenumber), options.uuid))
     # Overwrite chapternames if tree is flat and ComicInfo.xml has bookmarks
     if not chapternames and options.chapters:
+        filelen = len(filelist)
```
axu2, line 516: Just use `len(filelist)` per comment below.

kindlecomicconverter/comic2ebook.py
```diff
@@ -609,7 +624,7 @@ def imgFileProcessing(work):
             if opt.forcepng and not opt.forcecolor:
                 img.quantizeImage()
             output.append(img.saveToDir())
-        return output
+        return (output, spreadadjust)
```
axu2, line 627: no need for parentheses to return a tuple

axu2:
Closed due to inactivity. Reopen as needed. May already be fixed by #620

PR #616, opened 2023-11-13
fix kobo file names with .

### LINKED ISSUE #613: Filename change in Kobo profiles
Hi, I've been encountering this issue where with an input of a file with spaces and hyphens, those are deleted in the output file. For example : "One Piece - Chapter 4.cbz" will be named "OnePieceChapter4.kepub.epub".

This happens with any Kobo profile set as an option, maybe the kepub conversion does this. With a Kindle profile it works well and the filename is preserved (but not kepub).

### LINKED ISSUE #615: Filename problem if contain "." in Kobo profiles
If a file name input contains the character "." , all characters after that will be skipped in the output file. 
For example: "Vol.01 Ch.001 - Shinmen Takezo.cbz" will be named "Vol.kepub.epub".

Organizing files is super difficult.  Renaming the origin files to not include "." is impossible because the number of chapters can be easily 300.

axu2:
Prebuilt binaries for normal user testing (also contains recent error logging changes):

Windows: https://github.com/axu2/kcc/actions/runs/6853834938
Mac: https://github.com/axu2/kcc/actions/runs/6853836566
Linux: https://github.com/axu2/kcc/actions/runs/6853837839

 
- Fixes #615 

@HamzaM
@vietvu94 
Can you verify if this fixes your issue?

Various characters are illegal in older EPUB readers, including spaces. https://www.w3.org/TR/epub-33/#sec-container-filenames

Code based on: https://stackoverflow.com/questions/12985456/replace-all-non-alphanumeric-characters-in-a-string

Before and after:

image

I do not have a kobo to test so there's no way for me to verify if this build causes other issues.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -19,6 +19,8 @@
 #
 
 import os
+import pathlib
+import re
 import subprocess
 import sys
 from argparse import ArgumentParser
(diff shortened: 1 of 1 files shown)


HamzaM:
Just checked, this is what I get with the input : Oda Eiichiro - One Piece - Chapter 1053 The New Emperors.cbz

python3 kcc-c2e.py -p KoC -m -u -s -d -f EPUB -> Oda_Eiichiro_One_Piece_Chapter_1053_The_New_Emperors.kepub.epub
python3 kcc-c2e.py -p K -m -u -s -d -f KV EPUB -> Oda Eiichiro - One Piece - Chapter 1053 The New Emperors.epub

Both are considered epubs, but the Kobo one is handled differently in terms of filename compared to the Kindle profile.

Also, I find the author/creator set to KCC in the metadata, is there option to bypass that ? Because it later on messes up the metadata and puts every manga under the author KCC.

hvmzx:
Hi, I am still getting the same behavior, instead of spaces I got underscores everywhere. It'd be nice for the filename not to change

PR #633, opened 2023-12-03
 fix p7zip-rar error and files with ' or " by using arg lists instead of strings

axu2:
Not using the slightly unsafe `shell=True` parameter will actually throw a `FileNotFoundError` if I call a command that doesn't exist (like if kindlegen or 7z aren't installed). Called out explicitly here: 

https://docs.python.org/3/library/subprocess.html#exceptions

So prior code in #581 actually crashed when 7z and kindlegen aren't installed during the initial check. Our code prevent selecting MOBI/zip formats afterward, so no further exceptions expected.

So added try: except blocks as needed during initial checks.


axu2:
Also removed some 7z utf-8 encodings since it caused crashes on some Windows.

Issue #641, opened 2023-12-11
Package missing in environment.yml (conda)


jaroslawjanas:
**Describe the bug**
A `natsort>8.4.0` package was added as a dependency.
The `environment.yml` needs to be updated to reflect that. 
 
**To Reproduce**
Steps to reproduce the behaviour:
1. Install anaconda
2. Create the environment `conda env create -f environment.yml`
3. Activate the environment `conda activate kcc`

**Expected behaviour**
It should run the application.

**Screenshots**
![image](https://github.com/ciromattia/kcc/assets/28691446/a34aaad0-4f9e-4cbc-b815-55bddf400fd5)

**Additional context**
The *environment.yml* was added to simplify installation for conda users.
It should reflect *requirements.txt*.

PR #643, opened 2023-12-11
Added natsort to the dependency list in environment.yml

jaroslawjanas:
Issue: #641 

This adds missing `natsort` to the list of dependencies in *environment.yml*.
Versioning is kept the same as in *requirements.txt*.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,6 +9,7 @@ dependencies:
   - python-slugify>=1.2.1
   - raven>=6.0.0
   - distro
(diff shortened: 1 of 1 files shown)


axu2:
@jaroslawjanas Do you want to add natsort[fast] instead?

jaroslawjanas:
> @jaroslawjanas Do you want to add `natsort[fast]` instead? (I've don't use conda.)

Sure. I just checked if it works, all good on my side.

PR #653, opened 2023-12-30
Fix missing dependencies in setup.py

inganault:
setup.py dependencies don't have "natsort" and "distro", which will break pipx (and possibly pip) installation.


axu2:
Thanks!

PR #656, opened 2023-12-31
hide subprocess consoles on Windows

VampiroMedicado:
Hide the subprocess that runs **kindlegen.exe** when the program is performing the **.mobi** creation, to prevent the user from losing focus on what they are doing while a conversion is being done.


axu2:
D'oh! Sorry for not noticing this Windows-only quirk.

This is probably due to me removing the `STDIN=PIPE` in

- #633 
- https://stackoverflow.com/questions/24455337/pyinstaller-on-windows-with-noconsole-simply-wont-work

VampiroMedicado:
> @VampiroMedicado D'oh! Sorry for not noticing this Windows-only quirk.
> 
> This is probably due to me removing the `stdin=PIPE` in
> 
> - #633 
> - https://stackoverflow.com/questions/24455337/pyinstaller-on-windows-with-noconsole-simply-wont-work
> 
> You want to try putting it back? I notice a console window opens at EVERY point I call subprocess.run, including 7z, etc.
> 
> You can generate an exe with 
> ```
> python setup.py build_binary
> ```

Seems like the removal of `Popen` is intentional #581.

What about using a utils function `def silent_subprocess` to handle this?

axu2:
@VampiroMedicado 

Subprocess.run and popen have the same args. Re-adding `stdin=PIPE` everywhere as an arg should work.

axu2:
Actually, the proper way to do this is to use the 
```
creationflags=CREATE_NO_WINDOW,
```
arg in subprocess.run

VampiroMedicado:
> Actually, the proper way to do this is to use the 
> ```
> creationflags=CREATE_NO_WINDOW,
> ```
> arg in subprocess.run

Yeah, it sounds better now.

I was checking the docs and found nothing about how `stdin=PIPE` might interact with Windows Shell, but I'm not near my computer to test it. 😅

I'll update the PR tomorrow.

VampiroMedicado:
> Actually, the proper way to do this is to use the 
> ```
> creationflags=CREATE_NO_WINDOW,
> ```
> arg in subprocess.run

Yeah, it sounds better now.

I was checking the docs and found nothing about how `stdin=PIPE` might interact with Windows Shell, but I'm not near my computer to test it. 😅

I'll update the PR tomorrow.

axu2:
> I was checking the docs and found nothing about how stdin=PIPE might interact with Windows Shell, but I'm not near my computer to test it. 😅

I was basing this on the stackoverflow answer, but it seems to not be true.

Also, the `CREATE_NO_WINDOW` constant is Windows only and throws an Import Error on mac.

VampiroMedicado:
> > I was checking the docs and found nothing about how stdin=PIPE might interact with Windows Shell, but I'm not near my computer to test it. 😅
> 
> I was basing this on the stackoverflow answer, but it seems to not be true.
> 
> Also, the `CREATE_NO_WINDOW` constant is Windows only and throws an Import Error on mac.
> https://docs.python.org/3/library/subprocess.html#windows-constants
> 
> ```
> ImportError: cannot import name 'CREATE_NO_WINDOW' from 'subprocess' (/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py)
> ```
> 
> > What about using a utils function def run_silent_subprocess to handle this?
> 
> Sounds good, it just needs to check for windows or not.
> It seems this issue stemmed from not using the unsafe `shell=True`...

The current solution works right in Mac? I only have a Windows desktop.

axu2:
Yes, Mac and Linux don't create extra console windows already. This is windows only issue.  Just write your code so it doesn't throw an import error on mac

VampiroMedicado:
Hi, @axu2 happy new year 🎉

I made some changes, `utils.py` has the helper function and replaced any ocurrence of `subprocess.run` with it.

kindlecomicconverter/utils.py
```diff
@@ -0,0 +1,10 @@
+import os
+import subprocess
+
+def run_subprocess_silent(command, **kwargs):
```
axu2, line 4: I would call this `subprocess_run_silent` to make it sounds similar to what it's wrapping.
I would also put it in the existing `shared.py`

kindlecomicconverter/utils.py
```diff
@@ -0,0 +1,10 @@
...
+    if (os.name == 'nt'):
+        startupinfo = subprocess.STARTUPINFO()
+        startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+        startupinfo.wShowWindow = subprocess.SW_HIDE
+        kwargs.setdefault('startupinfo', startupinfo)
```
axu2, line 9: Is there a particular reason to use the `startupinfo` arg instead of `creationflags=subprocess.CREATE_NO_WINDOW`
https://docs.python.org/3/library/subprocess.html#subprocess.CREATE_NO_WINDOW

axu2 (changes requested):
A few comments, I can confirm it runs on macOS. Thanks for the great work!

kindlecomicconverter/utils.py
VampiroMedicado: You're right, I forgot we are checking for the OS.

VampiroMedicado:
@axu2 changes done.

Tested on Windows without issues.

axu2:
Looks great, I'll include these fixes in the next release!

axu2 (approved):
Looks great, I'll include these fixes in the next release!

axu2:
I've decided to backport this fix to v5.6.5. V5.7.0 is in prerelease and gets fewer downloads than v5.6.5

PR #657, opened 2024-01-02
fix Shift+Click convert button to select output directory renaming

### LINKED ISSUE #652: Disable file renaming for files with the same extension when input and output folders are different
When input and output files have the same extension (CBZ to CBZ conversion for example) kcc automatically adds "kcc_" to the output file name. I understand why it's for, but this also happens when you choose different output folder via shift click.

VampiroMedicado:
- Fix #652 

Add missing options dictionary parameter to allow search on a different folder for files with the same name.

Reporter confirmed that the behaviour is now working correctly: https://github.com/ciromattia/kcc/issues/652#issuecomment-1874129713


axu2:
Looks good!
Edited the PR title to make it more clear to ordinary users when they see the changelog.
Edited the first post to link it to the issue with magic `fix` word with formatting like:

```
- fix #652
```

Issue #662, opened 2024-01-11
Unsupported directory structure cover


axu2:
**Describe the bug**
Conversion fails if file structure looks like

Vol1/
--cover.jpg
--ch1/
----000.jpg
----001.jpg
--ch2/
----010.jpg

There's a loose cover image among chapter folders. 

**Additional context**
Current workaround:
Move cover into first chapter folder or Delete cover

Possible fix: automatically move cover into first chapter folder somehow or flatten file structure removing all sub folders.

axu2:
Flatten the directory structure is what we want.

PR #723, opened 2024-07-28
Added Calibre HTML output cbz support

Morialkar:
Ensured the check for Calibre HTML export is non destructive and won't impact the rest of the code.

I had a bunch of CBZ that were basically wrapping the Calibre HTML export in a cbz file. Added support to extract the images and place the cover at the top to then ensure proper function of the rest of the conversion process. If it's not a calibre HTML format (contains only an HTML file with a random identifier and a folder called RANDOMID_files) this update will not touch anything.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -808,6 +808,26 @@ def sanitizePermissions(filetree):
             os.chmod(os.path.join(root, name), S_IWRITE | S_IREAD | S_IEXEC)
 
 
+def detectCalibreHTML(path):
+    if os.path.isdir(path):
+        for root, _, files in os.walk(path):
+            if len(files) == 1 and files[0].endswith('.html'):
+                htmlFile = os.path.join(root, files[0])
+                htmlFolder = os.path.join(root, files[0].replace('.html', '_files'))
+                if os.path.isdir(htmlFolder):
+                    imagesFolder = os.path.join(htmlFolder, 'OPS', 'images')
+                    if os.path.isdir(imagesFolder):
+                        cover = os.path.join(htmlFolder, 'cover.jpeg')
+                        if os.path.exists(cover):
+                            os.remove(htmlFile)
+                            for f in os.listdir(imagesFolder):
+                                move(os.path.join(imagesFolder, f), root)
+                            move(cover, os.path.join(root, '0000.jpg'))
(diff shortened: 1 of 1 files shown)


axu2:
(I am not familiar with Calibre.)

Are you saying that cbz files generated from calibre work weird in kcc?

What was the original format before calibre converted to cbz?

Morialkar:
I downloaded a bunch of CBZ (that I did not personally generate) that didn't work with KCC and once opened had this directory structure. 
CleanShot 2024-07-28 at 19 12 45

So I added support for it. I assume it's Calibre related due to the calibreHtmlOutBasicCss file

Morialkar:
Also I'm dealing with a small bug locally but I added another catch to handle another set that didn't include the "OPS" folder and had images in the root _files directory

axu2:
Is this a special case of this general issue?

- #662

NaomiGilbertE22:
I don't think that's related, since the directory structure here is more:

Vol1/
--4iu32i_files/
----cover.jpeg
----OPS/
------images/
--------001.jpg
--------002.jpg
--4iu32i.html

Where I turn that into

Vol1/
--000.jpg
--001.jpg
--002.jpg

for proper processing.

But I can work in a fix to handle the cover.jpeg outside the first chapter in my logic to make the PR more valuable if you want, no trouble for me :)

Morialkar:
I don't think that's related, since the directory structure here is more:

Vol1/
--4iu32i_files/
----cover.jpeg
----OPS/
------images/
--------001.jpg
--------002.jpg
--4iu32i.html

Where I turn that into

Vol1/
--000.jpg
--001.jpg
--002.jpg

for proper processing.

But I can work in a fix to handle the cover.jpeg outside the first chapter in my logic to make the PR more valuable if you want, no trouble for me :)

axu2:
I don't know about this PR, there's duplicated code.

And we don't know what process creates files in this structure, because these are not CBZ files. These look like EPUB files renamed to CBZ. Since you didn't personally generate them.

I feel that it would be more useful to create a function that 

1) Checks if the files directory structure is compatible with KCC. https://github.com/ciromattia/kcc/wiki/Automatic-output-splitting
2) If it's not, flatten the file structure. And give a info message in the GUI that a flattening occured.
3) Remove unsupported directory structure warning.

This way, it would be more generic, because you may run into another slightly different file structure like OPS vs noOPS.

PR #727, opened 2024-08-01
Replaced StrictVersion with packaging.version.parse for Python 3.12 compatibility

### LINKED ISSUE #618: distutils is deprecated with removal planned for Python 3.12
```
kcc-c2e.exe
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\Cirn09\AppData\Roaming\Python\Python312\Scripts\kcc-c2e.exe\__main__.py", line 4, in <module>
  File "C:\Users\Cirn09\AppData\Roaming\Python\Python312\site-packages\kindlecomicconverter\startup.py", line 24, in <module>
    from .shared import dependencyCheck
  File "C:\Users\Cirn09\AppData\Roaming\Python\Python312\site-packages\kindlecomicconverter\shared.py", line 24, in <module>
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
```
https://github.com/ciromattia/kcc/blob/7273ca25b8c56d39cf79967317951048cd826f8d/kindlecomicconverter/shared.py#L24C39-L24C39

> The entire distutils package is deprecated, to be removed in Python 3.12. Its functionality for specifying package builds has already been completely replaced by third-party packages setuptools and packaging, and most other commonly used APIs are available elsewhere in the standard library (such as [platform](https://docs.python.org/3.10/library/platform.html#module-platform), [shutil](https://docs.python.org/3.10/library/shutil.html#module-shutil), [subprocess](https://docs.python.org/3.10/library/subprocess.html#module-subprocess) or [sysconfig](https://docs.python.org/3.10/library/sysconfig.html#module-sysconfig)). There are no plans to migrate any other functionality from distutils, and applications that are using other functions should plan to make private copies of the code. Refer to [PEP 632](https://www.python.org/dev/peps/pep-0632) for discussion.

https://peps.python.org/pep-0632/#migration-advice
> For these modules or types, use the standards-defined Python Packaging Authority packages specified:
> - `distutils.version` — use the packaging package

morpheby:
Simple drop-in replacement for the `distutils.StrictVersion`. Should partly 

- fix #618.

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -31,7 +31,7 @@ from PySide6.QtCore import Qt
 from xml.sax.saxutils import escape
 from psutil import Process
 from copy import copy
-from distutils.version import StrictVersion
+from packaging import version
 from raven import Client
 from tempfile import gettempdir
 
@@ -146,9 +146,9 @@ class VersionThread(QtCore.QThread):
             latest_version = json_parser["tag_name"]
             latest_version = re.sub(r'^v', "", latest_version)
 
-            if ("b" not in __version__ and StrictVersion(latest_version) > StrictVersion(__version__)) \
+            if ("b" not in __version__ and version.parse(latest_version) > version.parse(__version__)) \
                     or ("b" in __version__
-                        and StrictVersion(latest_version) >= StrictVersion(re.sub(r'b.*', '', __version__))):
+                        and version.parse(latest_version) >= version.parse(re.sub(r'b.*', '', __version__))):
                 MW.addMessage.emit('<a href="' + html_url + '"><b>The new version is available!</b></a>', 'warning',
                                    False)
         except Exception:
@@ -851,7 +851,7 @@ class KCCGUI(KCC_ui.Ui_mainWindow):
             for line in versionCheck.stdout.splitlines():
                 if 'Amazon kindlegen' in line:
                     versionCheck = line.split('V')[1].split(' ')[0]
-                    if StrictVersion(versionCheck) < StrictVersion('2.9'):
+                    if version.parse(versionCheck) < version.parse('2.9'):
                         self.addMessage('Your <a href="https://www.amazon.com/b?node=23496309011">KindleGen</a>'
                                         ' is outdated! MOBI conversion might fail.', 'warning')
                     break
--- a/kindlecomicconverter/shared.py
+++ b/kindlecomicconverter/shared.py
@@ -22,7 +22,7 @@ import os
 from hashlib import md5
 from html.parser import HTMLParser
 import subprocess
-from distutils.version import StrictVersion
+from packaging import version
 from re import split
 import sys
 from traceback import format_tb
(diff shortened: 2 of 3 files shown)


axu2:
Shouldn't we use version.Version and not version.parse?

Also, we need to remove distutils entirely, not just StrictVersion. Id check out the partial commit linked in the issue

morpheby:
@axu2 Good point. I wasn’t really paying attention and just fixed what I needed to start the KCC from python. I’ll check and fix everything else next time I have some free time :) (hopefully soon)

morpheby:
As for `Version` initializer vs `parse`, according to documentation both seem equivalent and a matter of style.

axu2:
The rest of distutils is only used when calling `python setup.py build_binary`.

Also, I'd prefer using `Version`, `parse` literally just calls `Version`,

https://github.com/pypa/packaging/blob/main/src/packaging/version.py#L47

requirements.txt
```diff
@@ -7,3 +7,5 @@ raven>=6.0.0
 mozjpeg-lossless-optimization>=1.1.2
 natsort[fast]>=8.4.0
 distro>=1.8.0
+packaging==24.1
```
axu2, line 10: Follow the `>=` pattern as the rest of the requirements.

kindlecomicconverter/shared.py
```diff
@@ -22,7 +22,7 @@
 from hashlib import md5
 from html.parser import HTMLParser
 import subprocess
-from distutils.version import StrictVersion
+from packaging import version
```
axu2, line 25: ditto

kindlecomicconverter/KCC_gui.py
```diff
@@ -31,7 +31,7 @@
 from xml.sax.saxutils import escape
 from psutil import Process
 from copy import copy
-from distutils.version import StrictVersion
+from packaging import version
```
axu2, line 34: I prefer 

```
from packaging.version import Version
```

Fun fact, parse literally just calls Version. But Version looks cleaner than version.parse

https://github.com/pypa/packaging/blob/main/src/packaging/version.py#L47

kindlecomicconverter/KCC_gui.py
```diff
@@ -146,9 +146,9 @@ def run(self):
             latest_version = json_parser["tag_name"]
             latest_version = re.sub(r'^v', "", latest_version)
 
-            if ("b" not in __version__ and StrictVersion(latest_version) > StrictVersion(__version__)) \
+            if ("b" not in __version__ and version.parse(latest_version) > version.parse(__version__)) \
```
axu2, line 149: That way it's just `Version(latest_version) > Version(_version_)`

axu2 (changes requested):
In `setup.py`, you need to remove 

```python
import distutils.cmd
class BuildBinaryCommand(distutils.cmd.Command):
```

with 

```python
class BuildBinaryCommand(setuptools.Command):
```

And verify that `python setup.py build_binary` works.

PR #743, opened 2024-09-10
Attempt at fixing "de-dupe cover" breaking chapters

### LINKED ISSUE #742: Chapters are off by 1 Page if "De-dupe cover" is enabled
**Describe the bug**
When De-dupe cover is enabled and a cbz with chapter info is being converted, the resulting epub has wrong chapter information. For example if page 6 is supposed to have a chapter, in the converted epub its page 7 instead, and therefore starts one page late.

**To Reproduce**
1. add a cbz with chapters inside a comicinfo.xml file
2. enable De-dupe cover
3. set output to EPUB
4. click convert

**Expected behavior**
I expected the chapters to be on the correct page, not one page late :P

**Screenshots**
kcc_settings

 - OS: Windows 11
 - Device: not applicable

**Additional context**
This also causes a "list index out of range" error when a chapter is the very last page

bust4cap:
should fix #742 

(also fixing general page spread alignment)

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -373,55 +373,64 @@ def buildOPF(dstdir, title, filelist, cover=None):
     else:
         f.write("</manifest>\n<spine page-progression-direction=\"ltr\" toc=\"ncx\">\n")
         pageside = "left"
+    firstLoop = True
     if options.iskindle or options.supportSyntheticSpread:
         for entry in reflist:
-            if options.righttoleft:
-                if entry.endswith("-b"):
-                    f.write(
-                        "<itemref idref=\"page_%s\" %s/>\n" % (entry,
-                                                               pageSpreadProperty("right"))
-                    )
-                    pageside = "right"
-                elif entry.endswith("-c"):
-                    f.write(
-                        "<itemref idref=\"page_%s\" %s/>\n" % (entry,
-                                                               pageSpreadProperty("left"))
-                    )
-                    pageside = "right"
+            if options.dedupecover and firstLoop:
+                firstLoop = False
+                continue
+            else:
+                if options.righttoleft:
+                    if entry.endswith("-b"):
+                        f.write(
+                            "<itemref idref=\"page_%s\" %s/>\n" % (entry,
+                                                                   pageSpreadProperty("right"))
+                        )
+                        pageside = "right"
+                    elif entry.endswith("-c"):
+                        f.write(
+                            "<itemref idref=\"page_%s\" %s/>\n" % (entry,
+                                                                   pageSpreadProperty("left"))
+                        )
+                        pageside = "right"
+                    else:
+                        f.write(
+                            "<itemref idref=\"page_%s\" %s/>\n" % (entry,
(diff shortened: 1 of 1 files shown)


axu2:
Wouldn't it be simpler to just add a offset of 1 on this line?

https://github.com/ciromattia/kcc/blob/master/kindlecomicconverter/comic2ebook.py#L534

bust4cap:
> Wouldn't it be simpler to just add a offset of 1 on this line if the option is checked?
> 
> https://github.com/ciromattia/kcc/blob/master/kindlecomicconverter/comic2ebook.py#L534

maybe i did something wrong but doing it that way if the first page is a chapter (aka "cover" for example) its instead just a copy of the last chapter again

kcc_chapters

axu2:
Oh that's because the -1 index is the last item of a list in python.

e.g.

```python
long_list = [1,2,3]
print(long_list[-1])
```
prints 3

So some additional logic to skip the first chapter if its only one page. Or clamp it to zero if multiple pages.

bust4cap:
someone more capable than me should probably take this over if another approach is preferred^^

axu2:
Went ahead and opened a PR please review and test.

axu2:
The last PR fixed the bookmark drift (while also deleting first bookmark).

Could you explain on a high level what this code is doing?

bust4cap:
i will close this for now, cause at this point it would likely clash with the fix you implemented already.

this had 2 fixes in one. the first one was simply trying correcting the page order for spreads, where the original code would put the page following a spread on the wrong side. (this needs more fixing in general anyway, cause depending on which spread splitting option you chose, this is even more borked, both in the original code and in my fixing attempt, which only fixes some of the issues)
the other one was for the chapters being off with dedupecover. instead of duplicating the first image as a cover page, it just sets the first image as the cover and sets the metadata accordingly. the original dedupe code just removed the (duplicate) cover file after its been added but doesnt handle the resulting page number discrepancy (to my limited understanding anyway)

axu2:
I see, this PR was a little complicated to review since it had multiple fixes. I might revert my change later if neccessary.

Are you still interested in fixing this:

> the original code would put the page following a spread on the wrong side.

Pretty sure what we need to fix this is to assign kcc-a files the `rendition:page-spread-center` property like here:

https://idpf.org/epub/fxl/#ex4

KCC only does left and right for kcc-b and kcc-c files.

axu2:
@bust4cap OK made a couple PRs that might affect this

axu2:
@bust4cap the sync fork button on the GitHub website on your branch is the way to merge master without weird stuff appearing in files changed.

bust4cap:
im terrible with github, sorry.
anyway, page alignment is fixed with your pull requests in every way i could find, yes

axu2:
@bust4cap 

- ref #784 

Please test if this does what you expect, instead of deleting the cover just change the initial page alignment to shift rest of book over.

and is there any remaining problems from your original PR?

axu2:
> instead of duplicating the first image as a cover page, it just sets the first image as the cover and sets the metadata accordingly.

This would be a good feature for a separate PR, save some space for that duplicated image and potentially avoid subtle bugs. @bust4cap

PR #767, opened 2024-11-03
simplify cover upload on newer MTP based Kindles by replacing EBOK with PDOC

### LINKED ISSUE #508: Kindle Scribe Cover generation after mtp firmware
Would be nice to have option to manually upload covers after the fact for longer batch conversions without needing kindle plugged in. 

Additionally, Kindle Scribe uses MTP so cover upload no longer works.

https://www.mobileread.com/forums/showthread.php?t=355924&highlight=mtp

## Hacky workaround

TLDR use kcc normally with a special flash drive inserted. 

1) Plug in a flash drive that will pretend to be kindle. Must be fat32 or msdos fat
2) Create `system/thumbnails` and `documents` folders at root of flash drive. Based on https://github.com/ciromattia/kcc/blob/master/kindlecomicconverter/kindle.py#L42
3) Convert with KCC and thumbnails will be generated there.
4) Copy the thumbnails onto your Scribe in the appropriate location. On Mac use MTP client Android File Transfer. Remember, on mac/linux only 1 program can access mtp at a time, so any other book transfer software running will interfere
5) Copy your books AFTER the thumbnails to avoid the Kindle blanking them out.

To create a fake physical flash drive, refer to https://support.apple.com/guide/disk-utility/create-a-disk-image-dskutl11888/mac
and use msdos format.

image

On Windows
https://learn.microsoft.com/en-us/windows-server/storage/disk-management/manage-virtual-hard-disks
use **FAT** not FAT32. 100 mb should be good, default on mac.

### LINKED ISSUE #685: PDOC vs EBOK option
It's a little annoying to always ask to rate the book at the end, which doesn't happen with personal documents from Send to Kindle.

### LINKED ISSUE #765: Kindle CS 12 will not display covers
Running KCC 6.2.1
As per the wiki I tried converting from a cbr/cbz with the kindle plugged in and connected. I used the Kindle CS 12 profile and converted to MOBI/AZW3.
image

Tried direct upload, file still shows up on device with the default cover.
Also tried uploading the same file and also a different file that was done during the same conversion though Calibre. Calibre does show the cover, but on the device it's still the same default one.
Restarted device between "upload runs".

Sorry if it's a stupid post but.....am I missing something?

Thank you.

axu2:
- fix #765 
- fix #685 
- fix #508 

- builds off #764

Test builds: 

~~https://github.com/axu2/kcc/releases/tag/v6.2.1-pdoc~~

~~https://github.com/axu2/kcc/releases/tag/v6.2.1-comicrack-pdoc~~

https://github.com/axu2/kcc/releases/tag/v6.3.0

The Kindle Colorsoft has been reported to not expose the system/thumbnails folder for covers, so here's a workaround and pre-built exe.

It just replaces EBOK tag with PDOC, so Kindle reads the cover directly from the file. Please test and tell me if it has issues.

Also, should I remove the ASIN tag?

And does anyone remember when pdoc covers started being read by Kindles? @Zappy86

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -380,7 +380,7 @@ class WorkerThread(QtCore.QThread):
                                     except Exception:
                                         pass
                             MW.addMessage.emit('Processing MOBI files... <b>Done!</b>', 'info', True)
-                            k = kindle.Kindle()
+                            k = kindle.Kindle(options.profile)
                             if k.path and k.coverSupport:
                                 for item in outputPath:
                                     comic2ebook.options.covers[outputPath.index(item)][0].saveToKindle(
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1227,7 +1227,7 @@ def makeBook(source, qtgui=None):
                 print('Error: KindleGen failed to create MOBI!')
                 print(errors)
                 return filepath
-        k = kindle.Kindle()
+        k = kindle.Kindle(options.profile)
         if k.path and k.coverSupport:
             print("Kindle detected. Uploading covers...")
         for i in filepath:
@@ -1249,12 +1249,13 @@ def makeBook(source, qtgui=None):
 
 
 def makeMOBIFix(item, uuid):
+    is_pdoc = options.profile in image.ProfileData.ProfilesKindlePDOC.keys()
     if not options.keep_epub:
         os.remove(item)
     mobiPath = item.replace('.epub', '.mobi')
     move(mobiPath, mobiPath + '_toclean')
     try:
-        dualmetafix.DualMobiMetaFix(mobiPath + '_toclean', mobiPath, bytes(uuid, 'UTF-8'))
+        dualmetafix.DualMobiMetaFix(mobiPath + '_toclean', mobiPath, bytes(uuid, 'UTF-8'), is_pdoc)
         return [True]
     except Exception as err:
         return [False, format(err)]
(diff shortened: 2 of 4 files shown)


AcidWeb:
Please research this topic before blindly messing with headers. This affect a lot of things other than covers.

axu2:
@AcidWeb Is there a list anywhere of differences?

AcidWeb:
No. It was never properly documented but it impact how e-reader and kindlegen handle such file.

axu2:
I see. I'll continue testing the files generated with this change and asking for feedback from testers. All I know is that EBOK was required in the past for covers but that isn't true now.

AcidWeb:
If plan to switch to PDOC then you in general can drop entire code that modify headers and send cover thumbnails .

axu2:
Will need to leave that code in for older kindles that can only get covers for EBOK. On mobile read a user confirmed it's  still needed for kindle voyage on latest firmware.

axu2:
Will tentatively force pdoc for these models:

```
    ProfilesKindlePDOC = {
        'KO': ("Kindle Oasis 2/3/Paperwhite 12/Colorsoft 12", (1264, 1680), Palette16, 1.8),
        'K11': ("Kindle 11", (1072, 1448), Palette16, 1.8),
        'KPW5': ("Kindle Paperwhite 5/Signature Edition", (1236, 1648), Palette16, 1.8),
        'KS': ("Kindle Scribe", (1860, 2480), Palette16, 1.8),
    }
```

asilentdreamer:
Kindle 11th gen (2022) does not use MTP, its just 11th gen (2024). Will there be an option to distinguish between the two?

axu2:
@asilentdreamer is there any reason to distinguish? It should still be able to read cover from PDOC.

asilentdreamer:
I see. In that case there doesn't seem to be one.

PR #772, opened 2024-11-10
fix bookmark drift when dedupecover is checked

### LINKED ISSUE #742: Chapters are off by 1 Page if "De-dupe cover" is enabled

axu2:
- fix #742 

Please test @bust4cap

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -530,8 +530,21 @@ def buildEPUB(path, chapternames, tomenumber, ischunked):
         elif options.splitter == 2:
             diff_delta = 2
 
+        first_chapter_is_cover_only = False
+        if len(options.chapters) > 1:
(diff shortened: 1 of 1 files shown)


axu2:
OK, I see the behavior difference now, but I'll have to look into it further at another time.

for now, I greatly simplified the code to just skip the first bookmark if option is checked, and -1 from all the later ones.

PR #780, opened 2024-12-02
Add additional check for differently named Fedora-based distros

termdisc:
Bazzite is Fedora-based but fails the `distro.id() == 'fedora'` check because `distro.id()` resolves to "bazzite." I added an additional check to include `distro.like() == 'fedora'` for situations like this.

This needs to be formatted as an or statement because plain Fedora resolves `distro.like()` as a null string.

### DIFF
--- a/kindlecomicconverter/comicarchive.py
+++ b/kindlecomicconverter/comicarchive.py
@@ -42,7 +42,7 @@ class ComicArchive:
             ['7z', 'l', '-y', '-p1', self.filepath],
         ]
 
-        if distro.id() == 'fedora':
+        if distro.id() == 'fedora' or distro.like() == 'fedora':
             extraction_commands.append(
                 ['unrar', 'l', '-y', '-p1', self.filepath],
             )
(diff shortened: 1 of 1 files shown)


axu2:
Appreciated, is this related to Steam Deck or SteamOS? @termdisc

termdisc:
This is related to Bazzite, which does have a Deck-focused image, but I wasn’t using KCC on the Deck.

SteamOS would be Arch-based, so it shouldn’t be affected by this change.

termdisc:
@axu2 Adding `~/.local/bin/` would be a fair addition because that's a common place where one would put userspace binaries. It wouldn't strictly need to be added to PATH if you add manual searching there.

I'm not well-versed in Python, but it looks like [pathlib.Path.home](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home) can help resolve this with `Path.home() / ".local" / "bin"` so that we ensure that we capture the user name as part of the folder path.

axu2:
Drafted in

- #786

PR #786, opened 2024-12-21
linux search for kindlegen in ~/.local/bin for steam deck and more

axu2:
@termdisc 

Can you test, I don't have any linux. This is slightly modified macos code, so unsure if it modifies linux behavior in an unexpected way.

I also just added all the macos locations as well.

This PR is because /usr/local/bin is read only on steam deck. @darodi


catsout:
Could you add XDG_DATA_HOME to search path?
The flatpak version is using this.
https://github.com/flathub/io.github.ciromattia.kcc/blob/master/patch/flatpak.patch

Issue #741, opened 2024-09-08
Dark Mode (feature request)


jeffmcneill:
Either dark mode switcher or simply follow color scheme of OS. This for Debian Linux.

ctrleo:
I use Debian too! I think the colour scheme of the UI changes dependant on your gtk theme, for example, I am using [Dracula](https://draculatheme.com) and my KCC looks like this:
![kcc](https://github.com/user-attachments/assets/d7786a80-fa88-4c2a-9cd1-c6094d887366)
If you could let me know which desktop environment you're using or which theme is currently installed I'd probably be able to help :)

jeffmcneill:
Yes, mine looks like that, also. The Dark Mode part desired is that large white area with dark text. Inverting that would solve everything for me. BTW I use Debian stable with XFCE, and Adwaita-dark (Gtk2, Gtk2) and Gnome theme for icons. Again, it is just the large white area with non-white text that would be usefully non-glaring if dark or dark grey.

ctrleo:
I found a fix for it!! All I gotta do is fix the text colours and I'll get a pull request out probably in the next hour ^-^

![happy tay](https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExY3p1enFqd2t6eHJ0NGpqbWp3a3I5NTllY2ZndGdwaGR3ZTlzZ2NmMyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/gromG2nn1K3Tg50ojX/giphy.gif)

PR #792, opened 2024-12-30
Implement system theme (dark mode) and fix background image on windows.

ctrleo:
This pr implements feature request #741

## Changes
- ~~Removed `background` (color) property from `list_background`, which allows the app to use the system default colour instead~~
- Added two options for `background-color` in `list_background`, which alternate depending on theme
- Removed the white filter over `list_background.png` and replicated the effect by lowering the image opacity to ~~70%~~ 10%
- Moved original `list_background.png` and it's GIMP editor file to `list_background_old.png` and `list_background_old.xcf`
- Recompiled UI files with new image
- ~~Created a subroutine in `shared.py` to detect if system is in dark mode or not~~
- Determines whether system is in dark mode or not using `QApplication.styleHints().colorScheme()` 
- Added `KCC_7.0.0.spec`, a file generated through `setup.py build_binary` on windows, to .gitignore

## This fixes...
- Aesthetic cluttering while in dark mode
- List Background photo hidden in WIndows
- My eyes while converting all my manga to kindle format

#### Before
Screenshot 2024-12-29 204312
#### After
Screenshot 2024-12-30 104623

#### After (in light mode)
Screenshot 2024-12-30 104549

#### Before
linux-before
#### After
newnewkcc

Ty for considering my contribution! ^-^


axu2:
I may be wrong, but there's probably a qt property you can read that tells you if it's dark mode?

https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5

ctrleo:
> I may be wrong, but there's probably a qt property you can read that tells you if it's dark mode? @ctrleo
> 
> https://www.qt.io/blog/dark-mode-on-windows-11-with-qt-6.5

omg thanks sm that would be way more efficient, I'll look into implementing it so that I can cut out the darkMode function

ctrleo:
I feel like the image blends too well with the text, so I'm gonna lower the opacity a little more

axu2:
I think it might be better to just remove the background image entirely in dark mode.

ctrleo:
@axu2 its kinda cute tho :( I'm trying to think of a way we might be able to pop a grey tint over it during dark mode which would make the letters stand out more

axu2:
Maybe make the image grayscale or black and white, maybe invert the colors, etc

ctrleo:
> Maybe make the image grayscale or black and white, maybe invert the colors on top, etc

thats an intuitive solution but it would be a challenge since it's difficult to change the image itself since it's compiled in with the ui files. I thought about putting a low opacity grey border around the text to help it stand out, which looks like this:
Screenshot 2024-12-30 030607
Screenshot 2024-12-30 030648
lmk what you think and i'll push the commit

axu2:
Just ran your code. Here's what it looks like on macOS. I'm personally against making the logo so prominent, I like it's subtlety.

I prefer (original no changes):

image


over (your changes):

image

Also, looks like the surrounding box and background changed colors.

For reference:

image


image

ctrleo:
> Just ran your code. Here's what it looks like on macOS. I'm personally against making the logo so prominent at least on light mode, I like it's subtlety.
> 
> I prefer (original no changes):
> 
> image
> 
> over (your changes):
> 
> image
> 
> Also, looks like the surrounding box and background changed colors.
> 
> For reference:
> 
> image
> 
> image

I do see it and I'm trying to implement some kind of mask for the logo while keeping a dark & light mode, but I can't help but notice that the text in the last image hasn't changed to white like it should have which is unusual since that bug hasn't come up in any of my testing.

ctrleo:
a little more progress
I lowered the opacity down to 30%, and I think its as close as im gonna get for subtle, heres the results:

Screenshot 2024-12-30 034105
Screenshot 2024-12-30 034124

the reason the current logo looks the way it does is because of a white cast filter over the image which essentially blends it with the background, however, when system dark mode is activated, gives us this:

Screenshot 2024-12-30 034303

(For reference, the text does display the correct colours, but there's a bug which causes it to default to one of the two colours in a sceenshot)

ctrleo:
I do think that helped, here's my changes to the UI essentially finished:
Screenshot 2024-12-30 040141
Screenshot 2024-12-30 035808
Screenshot 2024-12-30 040028
Screenshot 2024-12-30 040106

axu2:
This is getting there, but I want it to look like the example in the QT link earlier:

image

e.g. the text box has a white background in light mode, etc. currently the text box has the same background color as the window.

I can look into this later, this command may help (or just search for string "fusion"):

`git grep -In fusion`

ctrleo:
@axu2 gm! I looked into fusion and found that, with WIndows, it really takes that odd little accent colour feature and runs with it, see below:

Screenshot 2024-12-30 101248

Imo I feel like it just makes it look unnecessarily tacky, however, I'm gonna change the opacity of the logo again then I'll figure out how to brighten/darken the job list box compared to background.

ctrleo:
I brought it down to 10%, which I think gets us quite close to what it looked like in the original UI.

Screenshot 2024-12-30 102453
Screenshot 2024-12-30 102405

ctrleo:
> e.g. the text box has a white background in light mode, etc. currently the text box has the same background color as the window. And you could probably make the logo even more transparent.

I added back the background color property in `KCC_ui.py` and changed it depending on light or dark mode, heres my results:

Screenshot 2024-12-30 104623
Screenshot 2024-12-30 104549

on light mode it doesnt make much difference on a low brightness, I also added opacity to the background-color attribute, because the background-image tends to get stuck behind it on linux

axu2:
I appreciate your efforts, and it let to the temporary fix of dark mode via removing the background image entirely in KCC 7.1

But I'm just not sure if it's possible to preserve the original background color in light mode when adding a background image.


https://github.com/user-attachments/assets/06397ace-4e57-4782-8f7d-d7b88a6923a7

PR #709, opened 2024-06-18
A better page number-crop algorithm. (Manga)

neyney10:
## Summary
1. I wrote a new method to crop pages, as the current one fails.
2. Attached some cool gifs.

## Motivation
E-reader screen real estate is an important resource.
More available screen size means more details can be better seen, especially text.
Text is one of the most important elements that need to be clearly readable on e-readers,
which mostly are smaller devices where the need to zoom is unwanted.

By cropping the page number on the bottom of the page, 2%-5% of the page height can be regained
that allows us to upscale the image even more.
- Most of the time the screen height is the limiting factor in upscaling, rather than its width.

## The Current Algorithm Fails
While it is possible to adjust the 'power' of the cropping, I can only gain some success when the power is at least 4 (using the power user EXE), where the GUI itself only allows up to 2.

The current algorithm fails in all of the cases that I have at hand.
- I also don't want to crop the page number In case it would crop some manga-content.  Some high power levels cropped some of my pages too much.
- But I allow some level of cropping to achieve a tighter bounding box around the main manga-content with higher power levels.

## A Better Algorithm
I wrote a Python function using PIL, NumPy (openCV might perform faster, but I regressed to PIL as it exists as a dependency already and including OpenCV is heavy, package weight-wise).

<details>
  <summary> OUTPUT EXAMPLES OF THE NEW ALGORITHM </summary>

**(I extended the power level limit to 3, from the previous limit of 2)**

###  Page Number Cropping

ex1h001
ex5ga002

### Showcasing Its Ability To Retain Manga Content
**It does not crop the page number if it detects other things that might be cropped along with the page number.**

ex8n003
ex2h010
ex3b001
ex4bl001
ex6h006
ex7h009

</details>

## Issues
- My new algorithm uses the 'power' parameter differently than the other algorithm that simply crops the margins without the page number. A 'power level' in "Crop margins"  is different than a 'power level' in "Crop margins & page numbers". This might lead to inconsistent expectations from the user and confuse them.

PR #793, opened 2024-12-31
Fix cropping power sliders

axu2:
For some reason, Qt creator will delete several lines from the .ui file after 

- #709 

was commited. This messed up the cropping power sliders. So this is putting it back to how it was originally.

@neyney10 The only UI change should have been changing the UI slider maximum from 2 to 3, do you remember how those extra changes snuck in there? It was probably a Qt6 version mismatch though. I need to enforce the version.

I fixed this file with a `git checkout c35dd137eaca8f3e006f02d52f2167c4b3375dd9 gui/KCC.ui`


neyney10:
> For some reason, Qt creator will delete several lines from the .ui file after
> 
>     * [A better page number-crop algorithm with NumPy. (Manga) #709](https://github.com/ciromattia/kcc/pull/709)
> 
> 
> was commited. This messed up the cropping power sliders. So this is putting it back to how it was originally.
> 
> @neyney10 The only UI change should have been changing the UI slider maximum from 2 to 3, do you remember how those extra changes snuck in there? It was probably a Qt6 version mismatch though. I need to enforce the version.
> 
> I fixed this file with a `git checkout c35dd137eaca8f3e006f02d52f2167c4b3375dd9 gui/KCC.ui`

Yes I noticed them too. i didn't understand why the width changed while I didn't make any width changes and it doesn't seem to display them any differently on my end, so I figured its okay. i assume it is because I used a newer version of Qt or something.

PR #795, opened 2024-12-31
build appimage on ubuntu-20.04 instead of 22.04

axu2:
@jeffmcneill


jeffmcneill:
Tried to run this newest AppImage and got:

`Segmentation fault`

No other information.

Issue #804, opened 2025-01-07
[Feature] Inter-panel White Space Cropping


WirlyWirly:
Hello,

I'm coming to KCC as a long-time user of [MangAi](https://www.bitcake.com/projects/mangai/), which unfortunately has been EOL since the developer developed a sickness.

Therefore, I've found KCC and so far it's been working great! However, I don't see an option to perform inter-panel cropping to remove white-space.

As an example, I've attached a few comparison images; The first is the original image, secondly the KCC output, and finally the MangAi output with an example of horizontal white-space cropping.

Device: Kobo Forma (1440x1920)

*While the images appear to be the same size, this is only because GitHub stretches them to fit your display. Only the third image is actually representative of what you'd see on the screen of an eReader (Kobo Forma @ 1440x1920)*

Original (2138x3038)
![original](https://github.com/user-attachments/assets/73fb98d2-48e2-4443-9e36-7d286565146e)

KCC (1285x1920) | CroppingPower=3.0
![kcc](https://github.com/user-attachments/assets/256d231f-377c-4897-bfb3-b18282a9dddb)

MangAi (1440x1920) | Horizontal cropping, no option to remove page numbers
![mangai](https://github.com/user-attachments/assets/911dbf7c-076d-4ea8-b4a5-b741c9a68c94)

Notice that the panels in the third image have been horizontally cropped of white-space, a small threshold was left so that the images don't touch, and finally it was all blown-up as needed to fill the screen (while maintaining the correct aspect-ratio).

MangAi will also perform vertical white-space cropping, not just the margin white-space. Meaning it essentially passes over the entire image twice; First horizontally and then again vertically to remove all white-space. I couldn't find an image with vertical white-space between panels to use as an example, but it's essentially the same thing as horizontal cropping but done vertically.

While not a dramatic difference here, with certain pages it can definitely increase the used real-estate when all horizontal and vertical white-space has been removed... Also, MangAi uses ImageMagick, which seems to do a sharper job of resizing, but that's just me...

Thoughts?

neyney10:
Sounds interesting.
@WirlyWirly I couldnt run the MangAi as it requires a license. How does it work exactly - do u have a way to control how much space from horizontal/vertical white lines to remove? Like, in your example, you have shown that the app reduced the distance between the panels, but still kept some distance - do u have a way to control this distance? And is it an important feature?

Also - how does it work while also cropping the margins of a page? Is there any unique behavior?

neyney10:
[...]

PR #810, opened 2025-01-16
Inter-panel cropping method.

neyney10:
- fix https://github.com/ciromattia/kcc/issues/804

Unchecked: Disabled.
Partially checked: Crops horizontally.
Fully Checked: Crops both horizontally and vertically.

Please verify @WirlyWirly 
(Can download EXE from my fork https://github.com/neyney10/kcc/releases/tag/v7.1.2-inter-panel-crop)
 
Maybe todos:
1. Update the readme with the new flag/argument.

Notes:
- The method supports horizontally, vertically, and both. But the GUI only supports horizontally and both.
- It excludes sections that touch the borders of the image (they would mostly be margins) in order for this method to be used without cropping margins if the user does not wish to do so, contrary to the gifs I've shared in https://github.com/ciromattia/kcc/issues/804


gui/KCC.ui
```diff
@@ -237,6 +237,16 @@
          </property>
         </widget>
        </item>
+       <item row="6" column="2">
```
axu2, line 240: Add a tooltip in Qt Creator so text shows up on hover. and add arg to readme, wonder if there's a way to modify readme to update automatically?

kindlecomicconverter/KCC_gui.py
```diff
@@ -242,6 +242,7 @@ def run(self):
         options.cropping = GUI.croppingBox.checkState().value
         if GUI.croppingBox.checkState() != Qt.CheckState.Unchecked:
             options.croppingp = float(GUI.croppingPowerValue)
+        options.interpanelcrop = GUI.interPanelCropBox.checkState().value
```
axu2, line 245: - ref #785 

You aren't saving the value on close.

Maybe this PR should be added as example of adding new checkbox in the readme?

axu2 (changes requested):

gui/KCC.ui
neyney10: Oh yeah, i added the tooltip multiple times but I had to revert due to my issues with Qt / Qt creator version and I guess I forgot to add it again at the end.

I also added the arg to the readme back then ^_^".

neyney10: There is a way to modify the readme automatically... We can put the output of kcc-2e help in the readme by locating the section in the readme to replace with the updated text.

PR #843, opened 2025-03-02
fix error messages that only say a single number

axu2:
Partially revert 

- https://github.com/ciromattia/kcc/commit/30b8770e34926f3d5ee6e55d36abbb1d5b1e354c

Was causing vague error messages like:

```
2
```

Instead of like:

```
[Errno 2] No such file or directory: 'blahblahblah'
```


axu2:
@AcidWeb Is raven sentry client even used anymore? Can I just remove it?

PR #845, opened 2025-03-02
huge speed optimization on HDD by removing md5

axu2:
@utopiafallen Actually I don't observe any performance increase here.

More details: 

- https://github.com/ciromattia/kcc/commit/3bb8cc77788f7822ceb77af0f9e01223c4a514f5

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -37,7 +37,7 @@ from packaging.version import Version
 from raven import Client
 from tempfile import gettempdir
 
-from .shared import HTMLStripper, sanitizeTrace, walkLevel, subprocess_run
+from .shared import HTMLStripper, available_archive_tools, sanitizeTrace, walkLevel, subprocess_run
 from . import __version__
 from . import comic2ebook
 from . import metadata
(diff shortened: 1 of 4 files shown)


utopiafallen:
This is a huge slowdown for `buildHTML` (5-10x slowdown for me) and performance impact scales with image file size (I'm using 400-600MB CBZ files for testing). I strongly advise taking this change or coming up with an alternative to checksumming _the entire image_ just for a key into a dictionary (hash the source image metadata or something).

> it simplify the code and is much more error resistant.

From the original commit, this statement is also incorrect, at least in the strict sense of "error resistant." If you want to index the image metadata dictionary on an MD5 checksum of the image file, you need to store more metadata somewhere to detect checksum collisions, which I didn't see happening anywhere in the code.

axu2:
@utopiafallen Do you happen to remember the absolute speedup in seconds? 

An I missing something or is it just an md5 of the filename, not the actual image? 

```python
filelist.append(buildHTML(dirpath, afile, os.path.join(dirpath, afile)))
...
def buildHTML(path, imgfile, imgfilepath):
    imgfilepath = md5Checksum(imgfilepath)
...
return [md5Checksum(self.targetPath), flags, self.orgPath]
```

Your other changes are already a 2x improvement for me in overall processing time of `makeBook()` (2 minutes to 1 minute) with a 1.2 GB file.

utopiafallen:
`def md5Checksum(fpath):
    with open(fpath, 'rb') as fh:
        m = md5()
        while True:
            data = fh.read(8192)
            if not data:
                break
            m.update(data)
        return m.hexdigest()`

`buildHTML` went from 30s to 3s for me.

axu2:
Whoops, my bad. I didn't look closely enough at the md5 function.

Uh, I guess md5 is just REALLY fast on my CPU. I'll look into this further then.

axu2:
@utopiafallen 

Added a perf counter. 
Yup. 2 sec to 0.2 sec with a 1.2 GB file. The smallest optimization of the bunch from my perspective, so I didn't notice it in the overall processing time. 

Out of curiosity, what CPU are you using? Are you using an hdd or an SSD?

utopiafallen:
CPU: Epyc 7532
Storage: Network folder (10GbE) to an SSD/HDD hybrid (Windows storage tiering). You can treat it as an HDD for performance purposes with ~120MB/s peak sequential read.

Checksum/hashing is usually IO bound so your MBP SSD is at least 10x faster than my network folder and is probably why you (and the original author) never noticed the performance cost.

AcidWeb:
For the original author, this was the first bigger project in Python :-) Most of the flow of this application is either poorly designed or too slow. One of the reasons I abandoned this project was because I didn't have the strength to rewrite this application practically from scratch :-)

utopiafallen:
No worries, we all started from somewhere, and now people want to keep
using what you made, like me! 😁

axu2:
@utopiafallen Small change to make it more readable, once you give me a LGTM I'll merge:

- https://github.com/ciromattia/kcc/pull/845/commits/17bbea3b27739ee17481b80768b333ddf2fccd5a

utopiafallen:
Yup, looks good to me.

PR #848, opened 2025-03-03
all image files have unique ordered names regardless of subdirectory

axu2:


axu2:
@KeePach

Does this help at all with 

- https://github.com/koreader/koreader/issues/11497

PR #849, opened 2025-03-05
 Delete .travis.yml

iwa:
Hello,

It seems that you pushed a secret key into your travis CI config
Seems it's no longer used, but afaik this is not a token you should push

Don't know if you still use Travis but just in case!


axu2:
We don't use travis anymore, so I'll delete it entirely.

AcidWeb:
That key is encrypted. This is a correct way to store it.

iwa:
okey, Google Cloud and Travis's docs were a bit cryptic about it, so i thougth i'd make a PR just in case!

Issue #850, opened 2025-03-06
KCC - Error when converting HakuNeko folders


Lukas-Holecy:
**Describe the bug**
Using KCC from source.
I'm trying to convert a manga downloaded using HakuNeko Nightly into 200MB EPUB files with these settings:

![Image](https://github.com/user-attachments/assets/651a1d5a-1c0d-4fe6-8044-fe765fe06d7c)

After the processing images stage, I'm getting this error:

![Image](https://github.com/user-attachments/assets/58144b1a-fbf6-45ab-b39c-807daf7efef5)

This seems to be happening for all mangas that needs to be split into multiple EPUB files, though I don't know if splitting into multiple epubs is the cause.

**To Reproduce**
1. Clone KCC from github.
2. Using Python 3.13.2 follow installation steps.
3. Download large manga using HaruNeko
4. Set options for conversion as specified in the image above.
5. 3. Start conversion.
**Expected behavior**
Conversion should be able to convert into multiple EPUB files.

**Screenshots**
![Image](https://github.com/user-attachments/assets/651a1d5a-1c0d-4fe6-8044-fe765fe06d7c)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Kindle Scribe

axu2:
@Lukas-Holecy Whoops, this is a bug with the HDD optimization. You can revert that change or use 7.2.3 in the meantime while I fix this.

axu2:
My first thought is using 

- #848

As the dictionary key

Maybe sanitize earlier?

axu2:
@Lukas-Holecy I've fixed it in PR, can you test?

Lukas-Holecy:
> [@Lukas-Holecy](https://github.com/Lukas-Holecy) I've fixed it in above PR, can you test?

Works fine now. I've approved the pr if that means something.

Lukas-Holecy:
@axu2 Sorry I jumped the gun with the praise.
Your fix works, but it has terrible performance when converting really large manga.
If the source folder gets into GBs, the Processing images step completely kills my pc and prevents me from doing anything.
I had this problem with previous versions too, but I just lived with it. Eventually the process goes through, but it can take hours.

After using my version of 7.3, I thought that it was caused by old version of python, but your version brought the problem back.
The problem seems to be usage of video memory. The processor usage is about 10-15%, system memory gets to about 1gb, but video memory goes to 100% and I can't do anything. just switching windows takes minutes.

[...]

PR #851, opened 2025-03-06
Bug/850/setting empty metadata

Lukas-Holecy:
Ignoring missing image metadata (https://github.com/ciromattia/kcc/issues/850)
During conversion, the code expected that every processed file has a
metadata object attached in imgMetadata dictionary. For some reason this
is not true for every file.
I've added code that adds empty metadata to every file where the
metadata is missing.
This is more of a workaround, so the conversion can finish, it would be
better to find out why the metadata is missing in the first place, but I
don't see that.
Since most images have empty metadata, I don't think we are missing any
information.

This fixes https://github.com/ciromattia/kcc/issues/850


Lukas-Holecy:
This has much better performance for me that axu2 fix for large manga, so I'm reopening.

PR #852, opened 2025-03-07
rename images before processing to fix splitting issues

### LINKED ISSUE #850: KCC - Error when converting HakuNeko folders

axu2:
- fix #850

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -802,6 +802,7 @@ def sanitizeTree(filetree):
         for name in files:
             splitname = os.path.splitext(name)
 
+            # 9999 page limit
             slugified = f'kcc-{page:04}'
             page += 1
 
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -19,7 +19,7 @@
 #
 
 import os
-import pathlib
+from pathlib import Path
 import re
 import sys
 from argparse import ArgumentParser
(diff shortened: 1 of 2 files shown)


Lukas-Holecy (approved):
I've tested the new version and it's working absolutely fine for me now. So good job!

I am not very skilled with python and I definitely don't know enough about this project to leave really valuable review, but I don't see anything super wrong with this.
I have one piddly comment.
I see you've done a bit of a refactoring while fixing the bug. I'm talking about changes like:

22 - import pathlib
22 + from pathlib import Path

and 

704 -  src = pathlib.Path(srcpath)
705 + src = Path(srcpath)

I try to limit myself to only fixing whatever is related to actual bug when doings fixes, just so the fix is cleaner in the history.

Other than this insignificant nag, I have nothing negative to say about this pr, so I'm approving this.

Lukas-Holecy:
While I have previously approved of this PR, I found out, that I have huge performance issues when converting large manga.

This seems to be related to video memory usage for some reason. Even though the fix fixes the bug itself, the performance is so bad, I had to switch to my workaround again. It makes my PC completely unusable while the conversion runs.

axu2:
@Lukas-Holecy please state what you mean by very large manga and any relevant PC specs. I tested with 1.2 gb files on 8 gb of ram.

axu2:
> I try to limit myself to only fixing whatever is related to actual bug when doings fixes, just so the fix is cleaner in the history.

Fair enough, I'll remove that pathlib import change. 

I'll merge this for now to return it to previous behavior, then work on performance.

PR #853, opened 2025-03-07
memory optimization: store metadata in filenames, not a global dict

axu2:
@Lukas-Holecy This should also fix your memory issues. Instead of metadata dicts, store the metadata in the filename. this matters because image processing is parallelized.

- fix https://github.com/ciromattia/kcc/issues/850#issuecomment-2706339742

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -556,7 +556,6 @@ def imgDirectoryProcessing(path):
     global workerPool, workerOutput
     workerPool = Pool(maxtasksperchild=100)
     workerOutput = []
-    options.imgOld = []
     work = []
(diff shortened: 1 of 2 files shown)


utopiafallen:
In theory, it should since the process consuming the old image is the same one deleting it.

FWIW, I recommend against putting metadata in the filename because you may exceed MAX_PATH_LEN on Windows and end up crashing with a mysterious "file inaccessible" error.

I also don't understand how the metadata dict would be the cause of excessive memory usage since the dicts are just string-to-string and a cap of 9999 images would prevent that dict from getting very large. I'm guessing the missing image metadata issue originally experienced by @Lukas-Holecy led to misbehavior/memory leak elsewhere in the application.

PR #866, opened 2025-03-13
New "MOBI + EPUB (200MB)" output option

### LINKED ISSUE #749: [FEATURE REQUEST] Add EPUB (200MB) + MOBI output option
This makes it easier for the same conversion to produce results that can be used on the Kindle both by SendToKindle and copied over the cable. The MOBIs will be generated from the same EPUB parts, so if the EPUB is split into three parts to comply with the size limit, three MOBI parts will be generated. I believe that this way it will not be necessary to modify much of the code. Thank you.

simesherbs:
Adds new option that is basically combines the existing "MOBI+EPUB" and "EPUB-200MB" options. 
- fix #749


axu2:
The filesize of the mobi doesn't matter, the 200 mb limit only matters for the epub via Send to Kindle.

simesherbs:
> The filesize of the mobi doesn't matter, the 200 mb limit only matters for the epub via Send to Kindle.
> 
> Also, add the magic "fix" word like in your last PR to the first post.
> 
> and be sure to click the "sync fork" on your branch on github.com

Going forward, should all PRs have "fix" in it?

axu2:
fix will autoclose the linked issue when merged.

looks great! thanks for the contribution.

kindlecomicconverter/KCC_gui.py
```diff
@@ -922,7 +922,8 @@ def __init__(self, kccapp, kccwindow):
             "CBZ": {'icon': 'CBZ', 'format': 'CBZ'},
             "EPUB (Calibre KFX)": {'icon': 'EPUB', 'format': 'KFX'},
             "MOBI + EPUB": {'icon': 'MOBI', 'format': 'MOBI+EPUB'},
-            "EPUB (200MB limit)": {'icon': 'EPUB', 'format': 'EPUB-200MB'}
+            "EPUB (200MB limit)": {'icon': 'EPUB', 'format': 'EPUB-200MB'},
+            "MOBI + EPUB (200MB limit)": {'icon': 'MOBI', 'format': 'MOBI+EPUB-200MB'}
```
axu2, line 926: add a trailing comma so future git diff is smaller.

axu2 (changes requested):

axu2:
added the trailing comma

axu2 (commented):
Unfortunately, send to kindle now converts to kfx, which has different requirements than mobi conversion, so this has to be removed.

axu2:
Actually, nevermind, I found a way to use same epub for both. This feature stays.

Issue #665, opened 2024-01-11
Random blank pages on Kinlde Scribe


knowncold:
**Describe the bug**
Hi, I'm using kcc to convert my comics to kindle fomart. It's perfect for my Kindle Voyage, but got an issue on Scribe.

I found some blank pages when flipping pages. It's not true blank, if I flip back to previous page, return to home page and enter again, the blank page will restore to normal. And it's not some fixed pages, it's random.

You can refer to this [video](https://www.youtube.com/shorts/CuGTPbEQJQg).

**Desktop (please complete the following information):**
 - OS: macOS
 - Device: Kindle Scribe v5.16.5

**To Reproduce**
I'm using kcc v5.6.4 on Mac OS and Kindle Scribe v5.16.5, with command:

```
python3 kcc-c2e.py ~/Downloads/tmp -p KS -m -u -r 0 -c 1 --whiteborders  --forcepng -o ~/Downloads/test -b 2 --format MOBI
```

- I've tested same source files to profiles Kindle Voyage and Kindle Scribe, only Scribe has this issue.
- Same source files to epub, mobi and sideload mobi, `sendtokindle` epub, both have this issue.

And when I `sendtokindle` epub, got a email from Amazon:

> Thank you for using the Send to Kindle service. We are reaching out to let you know that a document, sent to your Kindle at, contains incompatible elements for Kindle format conversion. We’ve still delivered the document so that you can read it with Kindle, but in its current format it won’t support the latest Kindle features for documents, including Kindle Scribe handwriting features.
> 
> To make document(s) compatible with Kindle format conversion:
> Split large tables into several smaller tables, align the direction of text in the tables, and remove drop-caps in the tables.
> Remove multi-media content such as audio, video, and SVG images.
> Remove mathematical equations.
> Remove images that contain linear gradients.
> Remove password protection or any other security settings.
> Remove any footnotes inside footnotes.
> Check the document can be opened on a computer.
> Make sure the document doesn’t have more than 8,000 pages.
> Try sending the document again via Send to Kindle for Web.

axu2:
Scribe is glitchy with png. Use jpg instead. 

In fact mobi conversion is slow with png. 


- related #533

knowncold:
It works much better without `--forcepng`, I flipped 200 pages quickly and only got one blank now. I was thinking jpeg will lose some quality so chose png. I will try GIF as well.

Thanks for your help and work!

axu2:
[...]

Issue #563, opened 2023-07-28
Cannot output MOBI file larger than 400MB


tsopeh:
I want to convert a manga (larger than 700MB) so I can transfer it to my Kindle Paperwhite. There are the two options for me; I can convert it to MOBI/AZW3 or to EPUB format. If I covert it to MOBI/AZW3, than the output gets split automatically because of the 400MB limitation, and this is something that I would **really** like to avoid. On the other hand, the problem with conversion to EPUB is that it introduces some unavoidable margins, when transferred via Calibre to Kindle, meaning that the content looks smaller than when converted to MOBI/AZW3.

If I understand correctly, "automatic split" was introduced for MOBI/AZW3 due to the limitation of KindleGen. Was this limitation surpassed in the latest KindleGen? Is there a possibility that the "Indeterminate" option can be added for the "Output split" parameter, which would ignore size limitation, and try to create a MOBI file as large as the source data without splitting?

Is there some other way in which I can get large manga packed into a single file that gets displayed fullscreen/without margins?

darodi:
@tsopeh 

> If I understand correctly, "automatic split" was introduced for MOBI/AZW3 due to the limitation of KindleGen. Was this limitation surpassed in the latest KindleGen?

Same limitation in the latest `KindleGen`. 
As said [here [KindleGen]](https://www.amazon.com/b?node=23972728011)

>KindleGen is no longer available for download. Please use Kindle Previewer to convert, preview, and validate your eBooks. Kindle Previewer provides the same functionality of KindleGen

`KindleGen` is now found a sub directory of `Kindle Previewer`, but might disappear in the future.
So it won't be updated.

Kindles partitions are FAT32 and have a limit of 4GiB  (4.294 GB)

tsopeh:
Hi @darodi, thanks for the reply and explanation.

Issue #533, opened 2023-06-03
png slow with mobi


axu2:
Should use gif instead of png

Since kindlegen converts png to gif which multiplies the processing time. As noted in

- #41

On that note should probably add an option to match output format to input, since a png source converted to jpg can actually make things bigger and vice versa. I bet that's what's happening in 
- #349

EDIT: 

Was as simple as adding this: 

```
                # self.image.info["transparency"] = None
                self.targetPath += '.gif'
                self.image.save(self.targetPath, 'GIF')
```

GIF is problematic on Scribe though.

- #665 

breaks page flip feature too.

axu2:
@AcidWeb just fyi, gif doesn't get reconverted by kindlegen. Kindlegen reconverts png to gif.

Issue #349, opened 2020-03-29
CBZ conversion is generating much bigger file sizes than originals


aur0n:
I'd trying to use this wonderful tool to just crop white spaces in my CBZ comic books. In the settings, I only check the "Color mode" to not lose colors in the process and convert to CBZ again.

The problem is that the generated file is much bigger than the original.
For example, with a 18.7Mb comic book (mostly black and white pages) I get in result a 56.8Mb, that's a lot. I expected less or almost same size, since it's supposed to just crop white spaces.

Is there any way to fix this?

Joshuajrodrigues:
saaammmeee

axu2:
I bet it's because your source images are png and you are converting to jpg. a good idea would be to match the output format to the input. Converting from lossless png to lossy jpg for manga can make things way bigger. Jpg to PNG too

axu2:
Using the smart PNG mode should fix it.

PR #884, opened 2025-03-27
Add split size controls to GUI

antopanav:
Adds a GUI widget to reflect the `--targetsize` option of `kcc-c2e.py`.

Here's a screenshot of the result:


axu2:
Out of curiosity, what's the use case for this feature? This feature was initially added in 

- #347 

but since then, Send to Kindle added a 200 mb limit to the website which KCC splits on already.

For USB transfer, KCC will split on the maximum size of kindlegen.

antopanav:
I had a few conversions ending up being splitted with a big 450 MB file and a small 10 MB one. I found out and have been using the targetsize option already built in the CLI version of KCC and so I thought to implement it in the GUI.
It should behave same as the CLI in the GUI.

Similar, if not, identical, issue here:
https://github.com/ciromattia/kcc/issues/563



> Nice, always useful to add CLI features to the GUI.
> 
> Out of curiosity, what's the main use case for this feature? This feature was initially added in
> 
> * [Added targetsize option #347](https://github.com/ciromattia/kcc/pull/347)
> 
> but since then, I belive Send to Kindle added a 200 mb limit to the website which KCC splits on already.
> 
> For USB transfer, KCC will split on the maximum size of kindlegen.
> 
> Also, this number needs to be maxsized to the kindlegen max in code somewhere.

I'm not sure what's the maxsize of kindlegen, happy to update it.
I have tested it with 1024 MB targetsize and that's what I put as max in the GUI.

axu2:
The kindlegen max is defined here:

https://github.com/ciromattia/kcc/blob/3882b6ce0a1ee11ab7951047d50d1361d5817b2c/kindlecomicconverter/comic2ebook.py#L871

axu2:
Whoa, I stand corrected.

I had previously assumed the 400 mb limit was due to kindlegen expanding that to a 650 mb mobi, which is the stated maximum of kindlegen.

But apparently that's not true (anymore?)

Just tried your PR.

image

OK, I'll look into this further. Maybe kindlegen raised the limit at some point?

antopanav:
>> I have tested it with 1024 MB targetsize and that's what I put as max in the GUI.

> How big was the intermediate epub and final mobi in that case?

image

So the limit is on the input given to kindlegen? If so, maybe that's why it worked in my case, as it produced a less than 650 MB epub file.

I'm about to test with a bigger file and report back

AcidWeb:
There are more factors for which the current limits were introduced. Kindles struggle in general with so big MOBI files.

axu2:
- https://github.com/ciromattia/kcc/issues/136

AcidWeb:
Newer Kindles are less affected by this issue, but the problem is still there: slower page-turning, decreased battery life, reader responsivity during longer reading sessions, general performance when the user is "deeper" in a larger book and sometimes even crashes of the entire reader.

Not only does file size affect it, but the number of pages as well. That's why webtoon files have an even lower limit.

antopanav:
> KindleGen might support bigger files but realistically they were never used in Kindle ecosystem. Also please remember that both MOBI format and KindleGen itself is deprecated and abandoned by Amazon.

I can confirm, managed to find out at least 700 MB epub files are supported by KindleGen.


I understand about the performance impact bigger files may have on Kindles, in my case I had a few files exceeding the limit only by a couple of MBs, so I thought it would be handy to have the option available in the GUI.

Normally it is unchecked, it's up to the user enabling it by knowing what they're doing

axu2:
I see, I want to add this feature, but want to add a warning to the GUI so users don't footgun themselves (and if they do, they know why.)

So just add something like "Warning: chunk size greater than default may cause performance/battery issues, especially on older devices". Preferably not in the tooltip, somewhere more visible like right next to the splitSizeBox. Make sure default value changes depending on if webtoon or not.

Also, I'd greatly prefer that you didn't put the `splitSizeBox` right next to the checkbox. I'd rather that box appear when you click the checkbox. 

You can accomplish this by either:

1) Manually adding this to the `.ui` file. There are examples there already.

```
      <property name="visible">
       <bool>false</bool>
      </property>
```

This is the old method that KCC uses. Or use the modern method by setting it invisible in the __init__

https://stackoverflow.com/questions/7821481/initially-hidden-control-in-qt-creator#comment103531217_7843341

If these changes are messy you may be better off with a fresh PR.

Other topics that may interest you:

1) Using mozJpeg will compress better than using Jpeg.
2) if your source is png, you may be better converting to gif. Gif and png are both lossless and may compress better than jpg. But kindlegen will convert png to gif slowly. So adding gif support to KCC might be interesting.

- #533 
- #349 

However, newer kindles have issues with gif, so this may not work 

- https://github.com/ciromattia/kcc/issues/665

antopanav:
Thanks for the detailed response and the other topics!
I have noticed the `visible` property, so I may stick to that way, to be in line with how it's currently implemented.

The warning could be shown in the general text area on top, right after the splitsize checkbox is selected, or could do both actually (next to the splitSizeBox shown below)!

axu2:
> The warning could be shown... (next to the splitSizeBox shown below)

I suspect that's the best way to do it, but I'll let you decide.

antopanav:
Here's a preview of the suggested changes when all the widgets are shown, let me know what you think or if any more changes needed.
Do you think the "Chunk size" checkbox is ok as the last item of the middle column?

axu2:
OK, that's too visible. Make it one line, maybe like this (no red/bold, use default formatting):

image

And put the longer explanation in the tooltip.

> Do you think the "Chunk size" checkbox is ok as the last item of the middle column?

Yes, I may have plans for left column.

antopanav:
Here's the result, included the tooltip too:


The checkbox tooltip as well:



>Maybe disable this option if webtoon box is checked.

Disabling totally without giving the possibility of manually enabling it when webtoon is checked? Just making sure i understood correctly

axu2:
> Disabling totally without giving the possibility of manually enabling it when webtoon is checked? Just making sure i understood correctly

You can either

1) disable it totally if webtoon box is checked.
2) change the default box value to 100 if webtoon box is checked.

antopanav:
Just pushed the updates, it should work as expected.
I preferred to disable the control when webtoon is checked

axu2:
Great, made a small tweak. Thanks for the great work.

image

PR #889, opened 2025-03-30
speed optimization: fix accidentally cropping pages twice

axu2:
@neyney10 This looks fine? I noticed the `maybeCrop` function was being called twice per image, this shaves a second per volume for me.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -613,7 +613,7 @@ def imgFileProcessing(work):
             img = image.ComicPage(opt, *i)
             if opt.cropping == 2 and not opt.webtoon:
                 img.cropPageNumber(opt.croppingp, opt.croppingm)
-            if opt.cropping == 1 0 and not opt.webtoon:
+            if opt.cropping == 1 and not opt.webtoon:
                 img.cropMargin(opt.croppingp, opt.croppingm)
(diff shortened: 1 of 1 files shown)


kindlecomicconverter/comic2ebook.py
```diff
@@ -613,7 +613,7 @@ def imgFileProcessing(work):
             img = image.ComicPage(opt, *i)
             if opt.cropping == 2 and not opt.webtoon:
                 img.cropPageNumber(opt.croppingp, opt.croppingm)
-            if opt.cropping > 0 and not opt.webtoon:
+            if opt.cropping == 1 0 and not opt.webtoon:
```
neyney10: is "==1 0 and ..." intended?

axu2: Fixed already

neyney10:
I'll tell u what. I intentionally "double cropped" because the cropPageNumber method didn't crop at all if it couldn't find a page number (I even wrote at the function description that it should return "None" if not found). I see that I changed it since then and it actually does return a standard bbox without trying to crop a page number if it didn't detect one or found it being cropped with some manga-content.

So I guess its should be okay how you did it.

axu2:
I see, I thought it was unintentional due to the git blame. Thanks for background info, I didn't notice that since the code did what I expected.

image

PR #892, opened 2025-04-05
fix webtoon page numbers 

lalo101097:
I found that in the comic2panel.py file there is the following line: newPage.save(os.path.join(path, os.path.splitext(name)[0] + '-' + str(pageNumber) + '.png'), 'PNG') which causes that, if the name of the webton is not divided correctly, it uses the numbering 1,2,...,10 and that causes errors in some ereaders, if it is changed to: newPage.save(os.path.join(path, os.path.splitext(name)[0] + '-' + str(pageNumber).zfill(4) + '.png'), 'PNG') now it uses the numbering 0001,0001,...,0010 which does not cause those errors


axu2:
Thanks! The web toon code hasn't been touched in a long time and their are probably lots of issues with it.

PR #905, opened 2025-04-18
idea: UI changes

MirrorSim:
The request is about making a more intuitive UI, since it wasn't clear that drag & drop was possible.

For now, I put a placeholder, greyed out text which clearly explains the availability of drag and drop feature.

The text is always centered - thanks to `Qt.AlignmentFlag.AlignCenter` and only appears when the console window is empty.

I'd like to make the UI cleaner in the future.

--------------------------------------------------------------------------------------------------------------------------------------------

Old Title: multiple folders selection option
~~I've noticed it's not possible to select multiple folders at once - so in case of many folders to work with, the user has to select them one by one by himself, which takes lots of time. I guess this happens because the native (modern) Windows file dialog (folder selection screen) doesn't support multiple directory selection.~~

~~I've made a few changes to address that:~~

~~- When you have a parent folder with many subfolders (e.g. hundreds of manga‐chapter directories), holding CTRL [`Qt.KeyboardModifier.ControlModifier`] while clicking “Add image folder” now automatically enqueues only the subfolders—saving you from manually picking each one.~~

~~- A summary line (“n subfolders selected!” - where n is the number of subfolders selected) appears in the job list to confirm how many were added. This only happens when the user does the CTRL process.~~

~~- Added a line to the button’s tooltip to tell users this trick exists.~~


axu2:
What's your thoughts on highlighting multiple folders in windows explorer and dragging them into the kcc window?

MirrorSim:
my bad, didn't know this was possible 😅

Maybe we could make this more intuitive - like a greyed out font at the center of the kcc window when no folders are selected which says something like "Drag & Drop".

axu2:
Yea, a UI change to make it more intuitive would be nice, I think it's a common confusion. I also edited last comment with details.

MirrorSim:
Could you clarify the mobi part? Mine seems to work as intended.

Anyway, the whole ctrl thing seems useless and overcomplicated at this point. A UI change alone would be enough, I can start looking into that 👍🏻

axu2:
If you have say 10 chapter folders. Most people want to combine into single file. But sounds like you want to output 10 files. Kcc supports both, so not important. 

I'll let you decide how to make the ui clearer.

MirrorSim:
I removed my previous edits and added a centered, placeholder text when the window is empty. See if you like it :)

By the way, I'd like to build a cleaner UI (without excessive changes), as I'm familiar and worked on modern UIs in the past.

Also, just putting this out there: what if - after further UI changes - we remove the "Add image folder" button? 

It would be just a less effective version of drag and drop, since it can't select more than a folder.

axu2:
Yea the current UI could definitely be improved, and I've definitely thought about removing the add folder button because drag and drop is better. Let me know how you imagine it'll look like.

PR #917, opened 2025-05-01
CBZ output: Don't add padding around images when upscale unchecked or when white background is checked

axu2:
@mergen3107 @KaMyKaSii

- related https://github.com/koreader/koreader/issues/9163#issuecomment-1146637205

Do you think this change would help koreader users?

Previously, if you used an (~800x1200) image smaller than the device (1860x2480), and **upscale is unchecked** in cbz output mode, it would add huge margins.

This code change disables the added padding and outputs the image at it's original resoltion. Since KCC doesn't upscale to the huge resolution, this should mean much smaller filesizes at the same quality compared to using the upscale option.

test builds here:

https://github.com/axu2/kcc/releases/tag/v7.4.1-mapaki-koreader

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -372,7 +372,7 @@ class ComicPage:
         else: # if image bigger than device resolution or smaller with upscaling
             if abs(ratio_image - ratio_device) < AUTO_CROP_THRESHOLD:
                 self.image = ImageOps.fit(self.image, self.size, method=method)
-            elif self.opt.format == 'CBZ' or self.opt.kfx:
+            elif (self.opt.format == 'CBZ' or self.opt.kfx) and not self.opt.white_borders:
                 self.image = ImageOps.pad(self.image, self.size, method=method, color=self.fill)
             else:
                 if self.kindle_scribe_azw3:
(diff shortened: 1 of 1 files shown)


mergen3107:
Yes, I would leave image size unpadded. This way, say, in KOReader I dan long tap on the image and manipulate it with Image Viewer (zoom, rotate). 

Not that I had this use case before, because I always do upscaling outside KCC first (using waifu-2x), then use KCC to scale down

axu2:
KCC also adds black or white padding in the downscale case too, depending on background color detection. Do you think the downscale case is problematic too? @mergen3107 @MrBlank

```python
        else: # if image bigger than device resolution or smaller with upscaling
            ...
            elif self.opt.format == 'CBZ':
                self.image = ImageOps.pad(self.image, self.size, method=method, color=self.fill)
```

mergen3107:
I think I never had an issue with black/white padding at all. It always worked well, guessing the color right.

axu2:
I see, no one minds the downscale case, so I'll keep it around for now.

In the EPUB/MOBI formats, the background color is added via HTML/CSS, not to the image. But for CBZ its added directly to image. I thought adding padding directly to image might cause issues, like for example 2 page landscape reading, since the padding added to fit the screen perfectly in portrait would look strange in 2 page landscape.

I know some cbz readers like mihon have cbz background filling.

axu2:
Ah wait, I can just not add padding if white background is selected. White background and no padding is the same visually.

PR #920, opened 2025-05-06
cbz output: don't change subfolder names

### LINKED ISSUE #914: Sanitisation of subfolders and files names issue. Problems on TOC generation.
When a CBZ file with chinese characters subfolders are ran in KCC, with output format as CBZ, the resulting CBZ's subdfolder names are sanitised to pinyin. Files are viewed on a Kobo Clara BW with KOReader. 

Image

The end results of this sanitisation is, instead of the expected Chinese TOC, I get a TOC with pinyin.

Image

While I understand that KCC sanitises folder and file names for safety and compatibility reasons, the KCC-converted CBZ still retains the original Chinese filename and opens fine on my device. So perhaps, there can be an option for users to choose to **_NOT_** sanitise folder names? That's very important since many readers use subfolder names for TOC generation. The images file names can be sanitised, no issue with that. 

There would be a scenario where a user may choose this option unknowingly and cause an issue with their devices like Kindles etc, so perhaps only allow this option to be available only if a user chooses `Other` for profile.

The easy solution in my case is to use an archive utility and renaming the subfolder to their original's. KOReader picked up the TOC and displayed them just fine.

Image
Image

orpheus1120:
Do not slugify folder names when output format is CBZ 

- (fixes #914)

**What it does**

Early‑return in `slugify()` when `options.format == 'CBZ'`, so KCC keeps the
original (Unicode) folder names instead of transliterating them.

```python
def slugify(value):
    if options.format == 'CBZ':
        return value
    # existing code…

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -965,6 +965,8 @@ def createNewTome(parent):
 
 
 def slugify(value):
+    if options.format == 'CBZ':
+        return value
     value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
     value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
     return value

### CODE REVIEW

**Comment by axu2** [comment]
What's your opinion on only skipping the 

```python
value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
```

in the cbz case instead?

This way we would preserve the zero padding of numbers (e.g. 1 -> 0001)

```python
value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
```

Some cbz reader might sort 1,2,3,4,5,...10,11,12 as 1,10,11,2,3,4. I believe even koreader did that until recently.

orpheus1120:
@axu2 Honestly speaking, the right way is to pad both the image filenames and the folders, which also is the default behaviour in KCC. That's because we can’t really know which reader the user is using, and each one has its own way of interpreting file and folder order. So the safest approach is to ensure everything is alphabetically ordered using `slugify`, so the correct reading flow is guaranteed.

This PR is based on one key assumption: **that all folders and image names have already been padded before the KCC conversion**, since this practically skips the padding function. It works for me only because I’ve manually done all the padding beforehand. But that’s just my situation.

Now, are most KCC users also using Calibre? I would think so. If that’s the case, maybe we can expect them to organise the folders and files properly first, and only depend on KCC for the image conversion. If we go with that thinking, then we can make certain assumptions — like how in the comic-scanning community, it’s standard practice to use numeric prefixes. So maybe we can count on them to pad the folders too? How we answer this will affect how complex the code needs to be.

Your suggestion of using 

`value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')` 

and 

`value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))`

will work in most cases, but for mixed unicode and numbers like `第1话` (in chapters and TOC example in #914) - padding regex matches the 1 and converts to `0001` → `第0001话`. So it's not 100% foolproof and would still require manual intervention.

I have some ideas on how to improve this PR - like keeping all the Unicode characters as they are (so no transliteration when the format is CBZ), and only applying numeric padding to files that are obviously image files. But I’ve never done a PR before or contributed to an open source project, so I’ll need to read through the KCC code a bit, try writing some code, and test to make sure everything works properly first.

Thanks.

axu2:
Just to be clear: 

Original code (with comments):

```python
# only called on subfolders
def slugify(value):
    # convert all unicode to ascii via slugify
    value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
    # pad zeros to numbers
    value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
    return value
```

your current PR:

```python
def slugify(value):
    if options.format == 'CBZ':
        return value
    value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
    value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
    return value
```

I was asking for opinion on this alternative way to do it:

```python
def slugify(value):
    if options.format != 'CBZ':
        value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
    value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
    return value
```


**This code is only run on **subfolder** names.** The slugify_ext replaces the chinese characters with pinyin, the 2nd line pads the numbers of subfolders.

The image file themeselves are explicitly named `kcc-0001-kcc` elsewhere in the code. that zero padding is done elsewhere. Preserving the original filename caused order issues in the past, so completely replacing the image filename is needed. image renaming used to be handled by slugify function but not currently. image filenames have nothing to do with the slugify function

```python
def sanitizeTree(filetree):
    page = 1
    for root, dirs, files in os.walk(filetree):
        dirs.sort(key=OS_SORT_KEY)
        files.sort(key=OS_SORT_KEY)
        for name in files:
            splitname = os.path.splitext(name)

            # 9999 page limit
            slugified = f'kcc-{page:04}'
            page += 1
            newKey = os.path.join(root, slugified + splitname[1])
            key = os.path.join(root, name)
            if key != newKey:
                os.replace(key, newKey)
```

axu2:
> This PR is based on one key assumption: that all folders and image names have already been padded before the KCC conversion, since this practically skips the padding function. It works for me only because I’ve manually done all the padding beforehand. But that’s just my situation.

To be clear, you normally do this, but just didn't do it in this case? because padding I would expect 01,02,03...10.

image

> padding regex matches the 1 and converts to 0001 → 第0001话. So it's not 100% foolproof and would still require manual intervention.

To be clear, you don't like that behavior, and prefer it untouched? That's making the assumption the cbz reader will sort the subfolder correctly.

orpheus1120:
@axu2 Sorry for the confusion earlier. I took another look at the code block you proposed, and ran a quick test with the regex:

| Original name | Result |
|---------------|--------|
| `第1话`        | `第0001话` |
| `01 序章`      | `0001 序章` |
| `页面1234`     | `页面1234` |

So for CBZ we now keep unicode intact but zero-pad any 1-to-3-digit group for ASCII-sorting readers.  

Although visually I must say `第0001话` / `0001 序章` / `0010 第十回` look a bit odd in Chinese, but compatibility is key. The current PR works best for my case which I can always keep a copy for myself, no problem.

If you’re happy with what you proposed, I can update the PR to your preferred guard. Or better still you can submit the PR yourself.

> To be clear, you don't like that behavior, and prefer it untouched? That's making the assumption the cbz reader will sort the subfolder correctly.

Correct. I read on KOReader, and since 2023 i think it uses natural sort which can sort the folders in the screenshot. Hence I didn't bother to pad it.  If I'm reading on other readers, I will pad them.

axu2:
I see, thanks for clarifying. Based on your feedback, looks like only two options should be available. 

1) the current maximum compatible behavior.
2) completely untouched like your current PR

Anything in between in unneccessary.

Do you have any interest in adding a checkbox? You can use Qt creator to edit the .ui file and an example checkbox PR is here: https://github.com/ciromattia/kcc?tab=readme-ov-file#install-from-source

orpheus1120:
Sure I'll have a look.

orpheus1120:
Many thanks again for your involvement in this PR.

axu2:
I imagine it would probably look something like this:

```python
def slugify(value):
    if <box is checked>:
        return value
    value = slugify_ext(value, regex_pattern=r'[^-a-z0-9_\.]+').strip('.')
    value = sub(r'0*([0-9]{4,})', r'\1', sub(r'([0-9]+)', r'0000\1', value, count=2))
    return value
```

> As far as this PR is concerned, your proposed change offers a middle ground for CJK characters. Rather than transliterating all unicode, the CJK are intact. Not the most ideal of course but a significant improvement over the current release nonetheless.

Feel free to add that middle ground as a third option with the box partially checked, but it's not needed.

axu2:
Were you still interested in adding a checkbox? A lot of people are thinking about adding GUI features recently, do you think a video version of the "how to add a checkbox" would be useful example?

you don't even need to install anything, you just need to run `pyside6-designer` in the same place you run `python kcc.py` and open `KCC.ui`. @orpheus1120 

EDIT: Honestly, I'm considering just merging this as is and waiting to see if anyone complains. People in the KCC discord are asking for this feature.

axu2:
On even further thought, I'm going to merge this as is. 

If someone is using a cbz reader that doesn't respect subfolder order, they should use the feature of flattening the structure, removing all sub folders. So it would be a flat list of files. 

Or they can manually rename.

PR #924, opened 2025-05-14
Ui tweaks

MirrorSim:
Hi, I opened another pull request since my previous branch was a little buggy.

This was my previous request:

> The request is about making a more intuitive UI, since it wasn't clear that drag & drop was possible.
> 
> For now, I put a placeholder, greyed out text which clearly explains the availability of drag and drop feature.
> 
> The text is always centered - thanks to Qt.AlignmentFlag.AlignCenter and only appears when the console window is empty.
> 
> I'd like to make the UI cleaner in the future.

As you suggested, I removed the welcome messages at the start and made the 'empty window' message a little clearer (I also added a books emoji).

Now it should all appear at the start.

mmmù


axu2:
Yea, issues like 

- #925 

are pretty common so I want to merge this change. I'll review it soon.

for now, here's what it looks like on macOS:

image

axu2:
This seems a little complicated. Would it be simpler to subclass QListItem and promote it in `pyside6-designer`?

https://forum.qt.io/topic/161481/list-placeholder
https://doc.qt.io/qt-5/designer-using-custom-widgets.html#promoting-widgets

MirrorSim:
It does sound better, so - If I understand correctly - everything would be handled in one custom widget class? I'll look into that

axu2:
Yes, that's right. @MirrorSim 

I didn't include a video but the no items text vanishes when you drag and drop items.

axu2:
@MirrorSim Did you need any help with this? Here's the video example in case it's useful:


https://github.com/user-attachments/assets/b7932a3a-2de1-4ab0-bf65-07c050e25107

```python
from PySide6.QtWidgets import QListWidget

from PySide6.QtGui import QPainter
from PySide6.QtCore import Qt

class MyListView(QListWidget):
    def paintEvent(self, event):
        if not self.model() or self.model().rowCount() == 0:
            p = QPainter(self.viewport())
            p.drawText(self.rect(), Qt.AlignmentFlag.AlignCenter, "No items")
        else:
            super().paintEvent(event)
```

You just need to promote the QListWidget to your subclass in `pyside6-designer`.

- #969

Though I've made some window changes that might be simpler...

MirrorSim:
Hi, sorry for the late response. 

I finally got it working! The only thing that bothers me is that the Custom Widget displays a really small text despite me trying to tweak it with QFont and QRect, but I'm looking for a solution, I'll commit when it's done

axu2:
Just a heads up this is what it looks for me right now: 
image

MirrorSim:
> Just a heads up this is what it looks for me right now: image

If you resize the window is the text always kept to the center?

axu2:
Yes it stays center

MirrorSim:
ok I think we can make the block a bit more 'tight'

MirrorSim:
btw - is the joblistwidget implementation correct?



note: I don't know why 'KCC_ui_editor.py' and 'KCC_rc.py' appear to be changed, I believe they were automatically regenerated by gen_ui_files.bat but contain no actual changes

axu2:
That's because you are running an older version of py side probably. And Mac vs windows will generate slightly different things but those don't matter.

axu2:
image

MirrorSim:
Looks fine on Windows, at least...

image

axu2:
Hmmm, I'll play around with the screen sizes. What resolution is your screen? 1080p?

axu2:
Also, do you have the original code that I can use as reference point?

It may have been better to make a new PR instead of overriding this one.

I've got to apologize, this task looks much more complicated than expected, didn't mean to give you that much work 😅 

Also, does it briefly flash for you when you click convert? Did that happen in your original implementation?

MirrorSim:
I don't know why but I had a backup :))

[kcc-master.zip](https://github.com/user-attachments/files/21014340/kcc-master.zip)



> I've got to apologize, this task looks much more complicated than expected, didn't mean to give you that much work 😅

Not a problem! I've actually been really interested in UI programming lately and I'm learning so much.

axu2:
I'm sorry, but it seems that the other pull request that was merged before you remade this one where it removed the "Add directory" button and added a message in the window to drag and drop seemed to have eliminated people asking the question.

Also, I have also added manga bundle announcements to the window, so unfortunately I cannot merge this at this time.

I appreciate your efforts though, maybe there's an even better UI we can envision.

PR #944, opened 2025-05-27
ComicInfo.xml Title checkbox

### LINKED ISSUE #939: Better Metadata handling
I think the app will benefit from having more metadata options.

Here are a few that I can think of:
- The ability to change the prefix for number and volume
  - i.e. (Volume -> Custom Prefix) `My Hero Academia V1.0` -> `My Hero Academia Vol 1.0`
- The ability to parse the number to be an integer if it is `.0`, and fill in with `0` if the amount of characters is below a predefined amount. 
  - `My Hero Academia #1.0` -> `My Hero Academia #1`
  - `My Hero Academia #1` -> `My Hero Academia #01`
- The ability to use the title of the file as the name instead of `series + volume + number`
  - This actually might solve all the problems I have, and be easier to implement, since I name my titles properly.

**Example**
I use calibre and there is a plugin that i use to inject metadata into the ComicInfo.xml. The plugin uses number as the series volume number, which in turn I get the name of the file to be [Name of series] #[number] (i.e. `My Hero Academia #1.0`)
I would be nice to have a input like default author to be able to change the prefix of either volume or number metadata fields. For my case, I would like to be able to change # to v (i.e. `My Hero Academia v1.0`)

jorge-castellon-jr:
- fix #939


axu2:
In case I was unclear, you should run the command `pyside6-designer` in the venv where you run `python kcc.py` to open the qt designer and edit the `KCC.ui` file there. @jorge-castellon-jr

axu2:
Ok, I plan on releasing KCC 7.5 soon, and I wanted this feature included, so I finished it for you. Thanks for the initial work!

@vinhtq115 just fyi

axu2:
Ok, I plan on releasing KCC 7.5 soon, and I wanted this feature included, so I finished it for you. Thanks for the initial work! @jorge-castellon-jr 

image



@vinhtq115 just fyi

PR #945, opened 2025-05-30
Other profile max width change to 2400 to support 2400x3200 resolution.

yaqinking:
For 13.3 inch 300PPI devices.


axu2:
That's crazy, what device you using with 300 PPI and 13.3"?

yaqinking:
> That's crazy, what device you using with 300 PPI and 13.3"? That sounds like dedicated 2 page landscape reading.

BOOX Tab X C. And use MangaJaNai to scale up to this high resolution.

PR #959, opened 2025-06-10
Added ability to combine multiple cbz files into one

### LINKED ISSUE #612: [Feature Request] Combine multiple CBZ files
**Option for cbz files to be converted and compiled into one epub output.** 

At the moment I unpack cbz's, rename the folder to the chapter number, compress the file back to a zip, then append .cbz. KCC then converts the new file and all is good. Sometimes I will have to do this 100-200+ times for single chapters not placed into volumes yet.

I have tinkered around with EpubMerge in Calibre but the resulting image on my Kobo Sage is altered and gets clipped on the right side and bottom.

Greatly appreciate all your work, thank you.

Edit more details:

Manga chapter CBZs are released as they are translated each week. My workflow is selecting all and unpacking which results in folders (chapters) of images. Sometimes these need to be renamed to not throw an error message in KCC. Selected chapter folders are dragged into a new single folder create a volume folder. New single volume folder is compressed into a single zip. Resulting file is run through KCC to build an epub.

Would be cool if, for example, ten CBZ files are dropped in and with the option checked KCC could spit out one single epub.

Silver0006:
Added a new function which takes every CBZ selected in a single convert and combines them into one.

fixes #612

### DIFF
--- a/gui/KCC.ui
+++ b/gui/KCC.ui
@@ -590,6 +590,16 @@
          </property>
         </widget>
        </item>
+       <item row="6" column="0">
+        <widget class="QCheckBox" name="fileFusionBox">
+         <property name="toolTip">
+          <string>Combines all selected files into a single file. (Helpful for combining chapters into volumes)</string>
+         </property>
+         <property name="text">
+          <string>File Fusion</string>
+         </property>
+        </widget>
+       </item>
        <item row="2" column="1">
         <widget class="QCheckBox" name="upscaleBox">
          <property name="toolTip">
--- a/gui/KCC.ui
+++ b/gui/KCC.ui
@@ -593,7 +593,7 @@
        <item row="6" column="0">
         <widget class="QCheckBox" name="fileFusionBox">
          <property name="toolTip">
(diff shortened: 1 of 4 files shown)


axu2:
You performed your merge from master incorrectly. You need to click "sync fork" on your branch on the GitHub website, not doing a `git merge`. Otherwise a bunch of junk appears in the "files changed" tab when reviewing.

I'm excited for this feature however, it's a popular request.

Add the magic words to the first post

```
- fix #XXX
```

to auto link this PR to the issues it closes

Silver0006:
Should be good

axu2:
That seems to have made the "files changed" tab even worse 😅 

image



Maybe just copy your changes to a new PR?

But I see that it only supports cbz output, I don't think I can merge that, what users want is final merged epub/mobi output, and that feature might affect how cbz output is done. Or is that feature not possible?

axu2:
To be clear, the ideal would that the file fusion would happen first before processing, so output format wouldn't matter. @Silver0006

Issue #935, opened 2025-05-25
GUI unclear: custom output directory


jorge-castellon-jr:
### Custom Output Location
I recent was struggling with big that is currently being fixed by @axu2, thank you by the way, that made me think of a good solution that 1) will fix my issue with external exfat drive and 2) is just nice to have if you have multiple different directories in your input but you want all the output to be in a different folder. I can also see the code using the output folder to be the location all the work happens, like where the temp files can be stored.

*Example*
I use calibre and I have a plugin that can inject my metadata into the cbz ComicInfo.xml. So I start my process to add a manga on calibre to download metadata and then inject/embed it into the cbz file(s). If you haven't used calibre, calibre puts each book/manga into it's own folder. I would like to be able to have KCC grab all my inputs but then output to a specific folder i define.

### Custom Metadata Override Prefixes
Like I mentioned above, I use calibre and the plugin i use to inject metadata uses number as the series volume number, which in turn I get the name of the file to be `[Name of series] #[number]` (i.e. `My Hero Academia #1.0`)
I would be nice to have a input like default author to be able to change the prefix of either volume or number metadata fields. For my case, I would like to be able to change `#` to `v` (i.e. `My Hero Academia v1.0`)

axu2:
Does shift clicking the convert buton to set the output directory help?

jorge-castellon-jr:
I'm currently converting some manga right now, so i can't check.

Are you asking because that is currently a feature? or are you asking to see if that is a good solution to implement?

Issue #901, opened 2025-04-17
subchapter support


twilight6232:
My hardware device is KPW5 singnature ,I turned on the Show covers on lock screen option.

My comic file is relatively large, so in order to avoid slicing the output, I first converted the format to epub and then used the kindlegen command line to convert to mobi.

Compared with files that are converted to MOBI using KCC, the cover of the epub cannot be displayed properly.

Is there any way to make the epub format file converted with the kindlegen command display the cover page properly?

axu2:
What kcc settings (both you tried with mobi/epub) screenshot?

What kindlegen command did you run?

@twilight6232 

How close to chunking is your final file?

twilight6232:
![Image](https://github.com/user-attachments/assets/500d168e-f2da-4828-b184-62083fe26706)

twilight6232:
我的问题简单说，相比于直接由KCC转化的mobi文件，自己用kindlegen生成的mobi漫画无法正常在锁屏上显示封面。我想问一下是KCC执行kindlegen命令有什么不同导致这样的结果吗？

axu2:
I was asking what filesize the final epub is.

Don't use disable processing, that disables a lot of what KCC does and can lead to issues.

Maybe uncheck upscale.

Use mozjpeg to get smaller files if you are close to splitting threshold.

twilight6232:
If Disable Progressing is selected, the output EPUB file is about 500M. In fact, it doesn't matter.

My problem is simply that compared to the mobi files directly converted by KCC, the mobi comics generated by kindlegen cannot display the cover on the lock screen properly. 

I would like to ask what command KCC executes to make the cover on the lock screen work properly.

Is it different to execute the kindlegen command or is there some other processing?

This is a nice feature introduced in 7.4.0

axu2:
This is the command kcc uses: https://github.com/search?q=repo%3Aciromattia%2Fkcc%20-dont_append_source&type=code

twilight6232:
Let me give an example of why the disable progress option is used. Source file 143M

![Image](https://github.com/user-attachments/assets/86f2b1f1-df6f-46f6-989f-50a7a08c81b1)

twilight6232:
Regarding why I didn't directly convert the source files to MOBI, this involves another issue with KCC.

If the following file structure is used to generate the directory

![Image](https://github.com/user-attachments/assets/484dab30-dd34-4f05-a84e-8ace19f128c7)
[...]

Issue #608, opened 2023-10-28
When using the GUI, Unable to select multiple directories with 'Add Directory' button


jwyker:
When in the 'Add Directory' popup to select a directory.
I'm unable to select more than one directory at a time, if I have multiple directories, I have to add them one by one instead of SHIFT clicking to highlight multiple directories.

axu2:
Does selecting multiple directories in file explorer and dragging them into the KCC window directly work?

PR #960, opened 2025-06-11
Added ability to combine multiple CBZ into one files

### LINKED ISSUE #612: [Feature Request] Combine multiple CBZ files

### LINKED ISSUE #761: [Feature Request] Group files into a single output
I have my manga by chapter, and it would be nice to be able to organize it a little better. When selecting multiple files for the queue, a toggle to “merge queue into single file” would be really helpful. Even more so if it would follow the 200mb restriction you can set.

Thanks!

Silver0006:
Added a new function which takes every CBZ selected in a single convert and combines them into one. I moved the code around so it will work with epub and mobi. Special thanks to @axu2, I never would have bothered with switching the order of combining and optimizing the files.

- fixes #612
- fixes #761


kindlecomicconverter/comic2ebook.py
```diff
@@ -1056,6 +1056,8 @@ def makeParser():
                                      "2: Consider every subdirectory as separate volume [Default=0]")
     output_options.add_argument("--spreadshift", action="store_true", dest="spreadshift", default=False,
                                 help="Shift first page to opposite side in landscape for spread alignment")
+    output_options.add_argument("--filefusion", action="store_true", dest="filefusion", default=False,
+                                help="Combines all selected files into a single file. (Helpful for combining chapters into volumes)")
```
axu2, line 1060: File Fusion should be GUI exclusive feature based on how it works.

The GUI and CLI both call `makeBook()` on each file. There's no way to pass the CLI a list of files.

kindlecomicconverter/comic2ebook.py
```diff
@@ -1222,6 +1224,58 @@ def checkPre(source):
         raise UserWarning("Target directory is not writable.")
 
 
+def makeFusion(sources, qtgui=None):
+    filepath = []
+    GUI = qtgui
+    start = perf_counter()
+    pageTracker : int = 0
+    combinePath : str = os.path.join(os.path.dirname(sources[0]), "combinationTemp")
```
axu2, line 1232: Should name the final output the same name as the first file. Maybe append something at the ended like `fused`.

kindlecomicconverter/comic2ebook.py
```diff
@@ -1222,6 +1224,58 @@ def checkPre(source):
...
+        
+    #Ouput
+    sanitizeTree(combinePath)
+    print("Creating CBZ file...")
+    for paths in filepath:
+        if os.path.isfile(paths):
+            os.remove(paths)
+        elif os.path.isdir(paths):
+            rmtree(paths)
+    filepath = makeZIP('Combined_comic', combinePath)
```
axu2, line 1261: zipping the final output is unneccessary, KCC can take as input nested folders of images. And the name should be the first source's file name, plus something at the end like `fused`

kindlecomicconverter/comic2ebook.py
```diff
@@ -1222,6 +1224,58 @@ def checkPre(source):
...
+    combinePath : str = os.path.join(os.path.dirname(sources[0]), "combinationTemp")
+    os.mkdir(combinePath)
+    print("Running Fusion")
+        
+    for source in sources:
+        print(f"Processing {source}...")
+        
+        checkPre(source)
+        print("Checking images...")
+        path : str = getWorkFolder(source)
```
axu2, line 1241: I think it would be much simpler to instead of calling `getWorkFolder()`, to instead write a simple function that is similar to `getWorkFolder()` that just moves/extracts all the images into one folder.

E.g.

If filefusion is checked:
1) create a temp folder same name as first source (the fusion folder)
2) move/extract all the files to be fused into individual folders inside the fusion folder

Example:
I start with `TITLE - ch1.cbz` and `TITLE - ch2.cbz`
I checked the fusion box.
KCC should create a **folder** called `TITLE -ch1-fused` with 2 subfolders inside with each chapter extracted. 

And then passes that folder through the normal process.

axu2 (changes requested):
Thanks for fixing the merge issues, reviewing is much easier now. I also merged in the latest changes from master to fix a bug from the dev branch.

gui/KCC.ui
```diff
@@ -590,6 +590,16 @@
...
+       <item row="6" column="0">
+        <widget class="QCheckBox" name="fileFusionBox">
+         <property name="toolTip">
+          <string>Combines all selected files into a single file. (Helpful for combining chapters into volumes.)</string>
+         </property>
+         <property name="text">
+          <string>File Fusion</string>
+         </property>
+        </widget>
+       </item>
```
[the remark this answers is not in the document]
Silver0006: Windows & pyside6-designer

axu2:
Looks great! Functionally, it's perfect. subfolders are preserved and become chapters in the final mobi/epub.

Just a small nitpick: the final output being named after the temporary directory with random characters at the end, but I can fix that if you are tired of the back and forth. I'll merge this after you/I fix that. I also plan to make the temporary fusion directory have a simpler name to avoid confusion.

Could you clarify what you mean by this:

> Fusion cannot select folders individually. You must select folder containing folders.

As a user, I selected all the folders/cbzs individually in bulk by highlighting them all, dragged them all into the KCC window, and clicked fuse. I didn't drag in a folder containing folders. that sounds like the opposite of your text.

(If you got confused by the video, what I care about is the final output, not the process of dragging in a folder containing sub folders.)

and is there any functionality you wanted that you think was lost by simplifying the code?

Silver0006:
Yeah, if you'd be willing to fix the file output that'd be great. Otherwise, I think everything should be good.

axu2:
@Silver0006 

Alright, simplified your code, please double check that it does everything you want it to.

Just some notes for future:

1) pathlib.Path is very powerful and can simplify things significantly (and using stem makes handling files/dir easily without distinction)
2) os.renames automatically prunes empty directories
3) subchapters are not supported, that would be separate PR, so I'm flattening any subfolders of subfolders. 

- #901 

I'll merge this once you double check.

Silver0006:
Adding files works fine, but I'm having some trouble combining image folders. Is the idea that people will combine image folders themselves? It also could be a difference in OS if you're not on windows. 

The concept I had was that it should be able to combine any input together. So I had the code loop through each input and extract them, then it would take the extracted images and make a single file output.

Using Add image folder
- Provide a single folder that would then be checked for other folders which would be used as chapters
- Select multiple image folders that would be combined into one with each folder acting as a chapter.

Using Add files
- Provide multiple files which would each represent a chapter.

My intention was not to do subchapters though it seems that's what I may have done.

With the changes that have been made selecting two folders separately is having some kind of failure with the directory `Fusion Failed. [WinError 183] Cannot create a file when that file already exists: 'C:\\Users\\Aidan\\Documents\\Mangas\\Discommunication Seireihen\\KCC-43v2zjew\\OEBPS\\Images' -> 'C:\\Users\\Aidan\\Documents\\Mangas\\Discommunication Seireihen\\Vol.01 Ch [fused]\\Vol.01 Ch'` Had this problem earlier so something in my code that accounted for this was deleted. You can see that the name of the directories are getting shortened to `\Vol.01 Ch [fused]\\Vol.01 Ch'`

Selecting one folder results in `Fusion Failed. Fusion requires at least 2 sources.` 

So I'm up for whatever you want to do!

axu2:
Right... I'm seriously considering removing the "add image folder" button from the GUI. The proper way is to drag folders into the KCC window, like here (this is only a appearance change, behavior is the same in current KCC):

- #924 

> Provide a single folder that would then be checked for other folders which would be used as chapters

This case is what KCC does when file fusion is unchecked. Guess I could add a note to uncheck in this case. But yea fusion should require 2 sources.

As for those cut off chapters, I see that things past the `.` are getting cut off. That is what `stem` is doing, I guess i can't use stem then. I'll fix that.

axu2:
@Silver0006 

Alright, that should do it. Btw, what's your opinion on that UI change?

Silver0006:
On the UI would it be difficult to change _Add image folder_ to work the same way _Add file(s)_ works? That way it has the same functionality as dragging the files onto the window.

axu2:
I don't think so, I think it's a Qt limitation. Dragging folders into the window is pretty easy anyways.

- https://github.com/ciromattia/kcc/issues/608

Silver0006:
Yeah seems like you'd have to build a custom button in Qt, so probably just better to drop it. I do think there need to be a better indicator to show you can drag and drop files. Maybe add a message that tells you to drag and drop using `MW.addMessage.emit()`?

axu2:
> Maybe add a message that tells you to drag and drop using MW.addMessage.emit()

that sounds like a nice simple solution, I support it. Want to open a separate PR for that? Make sure it mentions files AND folders.

And everything functionally works well with latest commits?

Another unclear GUI issue you might be interested in:

- #935

Silver0006:
Everything looks good. I'll add the message after the merge, so it can be in a separate pr. Potentially we replace the Add folders with a select output?

axu2:
merged.

I'll let you decide what works best in terms of the GUI, but I am predisposed to removing the Add folder button since it's simply not useful.

and yea, parallel GUI changes are tricky when code generation is involved, I wonder if there's a way to solve that. But KCC is pretty quiet, so that issue hasn't really come up.

- #964

axu2:
Oh, btw, just a survey, what device are you using for reading?

Silver0006:
Kobo Libra Colour

axu2:
Consider using the epub/kepub output of kcc or using koreader.

- https://github.com/koreader/koreader/issues/11877

axu2:
@Silver0006 

Adding a folder/file drag and drop message is probably just one line here: https://github.com/ciromattia/kcc/blob/ee375abfc51da95139bbc3e97a535a104a7154c0/kindlecomicconverter/KCC_gui.py#L1119

Silver0006:
@axu2 This week I'm going to work on replacing the add folders with a way to select a output directory

for https://github.com/ciromattia/kcc/issues/935

axu2:
@Silver0006 Ok, I already added another message to the window, I think that should fix that, and is just a simple message.

- #969

axu2:
@Silver0006 

I actually made a mistake, you actually can pass a space separate list of source folders/files to kcc on command line.

PR #975, opened 2025-06-17
Replaced Add Folders with output directory button.

### LINKED ISSUE #745: Inconsistencies with output directory in Windows 10
# Use case
**I like to keep my source and output files separate**, so I would prefer not having to shift-click the convert button and having to specify my output directory each and every time I start a conversion.

# Current inconsistent behaviour
- By default, when I click the "Convert" button (without holding shift) the output directory defaults to individually matching each source item.
- When I shift click the "Convert" button, I am prompted to select an output directory (for all items in my queue), **the starting point directory is always the same as application executable**.
- Each time I add files or directories to the queue by clicking the "Add file" or "Add directory" buttons, **the application changes its starting point directory** for the dialogue which appears when I shift-click the "Convert" button. The starting point directory will now match the source file's location (instead of the executable directory). This change persists after the application is closed.
- When I shift-click the convert button but choose "Cancel" in the output directory selection dialogue, **the app starts the conversion process regardless** (placing each output file in the directory that corresponds with its source file, as if I hadn't shift-clicked the "Convert" button in the first place.)

# My current workaround
1. Make a temporary copy of a random source file and place this in my desired output directory
2. Add this temporary file to the conversion queue using the "Add file" button in the application GUI (the "Add directory" button also works)
3. Clear the queue or close the application, and delete the temporary source file
4. Add any files I actually intend to convert using drag & drop into the application window, instead of using the "Add file" or "Add directory" buttons (since that would reset the starting point directory to be same as the source)
5. Shift-click the "Convert" button (the output directory selection prompt should now start in the directory where the temporary file was) and choose "Select folder" in the output directory selection prompt (the starting directory will now be where the temporary file was, rather than where the application executable is)

# Feature request
An optional setting for selecting a default output directory which will be used whenever i click the convert button normally (without needing to hold the shift key)

Silver0006:
I took the code from #969 and replaced the selectDir() function with it. Then replaced the Add Folders button and linked it to selectDir(). The button only works if the checkbox next to it is selected. The idea is you can set a output directory over multiple converts instead of re-selecting it with shift-click for each convert. Shift click for single converts still works

fixes #745


Silver0006:
Something got messed up with UI generation. Trying to fix it

Issue #932, opened 2025-05-25
[Feature Request] Color mode bit depth


Rinkerbel:
When color mode is turned on, all images are encoded to 24 bit depth. I like colors but I also value my disk space. Is it possible to have only the color pages encoded to 24 bit and grayscale to 8 bit?

axu2:
@Rinkerbel Implemented, test builds here, please try and feedback: https://github.com/axu2/kcc/releases/tag/v7.5.1-custom-output-folder

Rinkerbel:
I tested the build and it looks to work great. I did notice some quality loss at 4 bit tho. I'd prefer the grayscale images be 8 bit.
Here's a comparison for you to look at: https://slow.pics/c/Zh8IGIMS

I also loaded them onto my ereader (kobo sage) and I think I can see the difference between 4 and 8 bit? It is very minimal though and I may be halucinating things
In my testing, 4 bit yielded about a 2.5% reduction in filesize, personally I'd wanna keep it at 8 bit.

These are my kcc settings:
Current:
`kcc-c2e --profile KoS --manga-style --splitter 1 --cropping 0 --forcepng --delete --format CBZ $items`
With Color:
`C:\Users\Rinkerbel\Downloads\Programs\KCC_c2e_7.5.1.exe --profile KoS --manga-style --splitter 1 --cropping 0 --forcecolor --forcepng --delete --format CBZ $items`

The images processed by kcc were also 8bit images

axu2:
@9783e6 just fyi

axu2:
@Rinkerbel Wait, that quality loss thing might be a bug. Looking into it. I didn't notice since I use jpg myself.

axu2:
Nevermind, that bug I mentioned is unrelated. 

- #991

Could you try using the new and old kcc to create 4 and 8 bit versions. 

Then extract the images. 

Then put them into 1 folder, renaming so 4 and 8 bit images are next to each other for easy compariosn.

Then zip into 1 folder, and rename that zip to CBZ.

To create a file where a page turn shows the comparison on your Kobo? Comparing things on your PC screen mighg show differences invisible on eink. 

Why are you forcing color on a Kobo sage? Maybe set gamma 1.0 instead

axu2:
New test build: https://github.com/axu2/kcc/releases/tag/v7.5.1-png @Rinkerbel

Rinkerbel:
[...]

PR #976, opened 2025-06-17
Fix png bit depth

9783e6:
Currently all PNGs are always saved in 8-bit depth when the forcepng option is enabled and forcecolor is disabled, even though the image was previously quantized to 4, 15, or 16 grayscale shades (depending on the selected profile). The fix ensures that png images are saved with the appropriate lower bit depth in these cases.

In my tests, this change results in ~8% file size reduction when using the CBZ as output file format. (Also seems to improve page turning speed on my kindle in koreader).


axu2:
I haven't touched the bit depth part of the code or really thought about bit depth at all, could this improvement be applied to jpg as well?

9783e6:
Standard jpg only supports 8bit depth, so no, I don't think it's possible

axu2:
On that note, doesn't eink only have 16 shades of gray, so wouldn't 4 bit be enough in all cases instead of 8 bit? For example, eink carta 1200 on kindle scribe.

https://www.eink.com/product/detail/ED103KC2

> Number of Gray 16 Gray Level (monochrome)

@9783e6 

This might be of interest to you:

- #932

9783e6:
For grayscale eink, yes, 4bit should be enough in all cases. But for color eink, it's more complicated. Technically kaleido 3 is still supports only 16 levels of gray + 16 levels of red, green and blue (so, 4096 colors in total) so 4 bit should be enough overall, even for color eink (if we save 4bit per channel per pixel) but png's RGB image type (truecolor) only supports two bit depths 8bit and 16bit ( https://www.libpng.org/pub/png/book/chapter08.html#png.ch08.div.5.6 ) so we still need 8bit pngs if we want to save color

axu2:
Overall, looks good. But let me make sure I understand what you did. Basically:

1) You removed the code that expands the palette into 256 characters. This was probably a leftover from old PIL version since the code hasn't been touched in 13 years, it probably only worked on 256 long palettes back then.
2) You removed the conversions to RGB, also likely a leftover from old PIL versions when functions were only implemented for RGB.

Then finally the PNG encoder is smart about how many bits it needs when saving in the end?

9783e6:
Yes, that's exactly right, I removed what appeared to be legacy code. Pillow now handles palette images more robustly, and the save() function internally determines the appropriate bit depth based on the number of colors in the palette: https://github.com/python-pillow/Pillow/blob/ef0bab0c6579fd00987740a6a327603ff3886a38/src/PIL/PngImagePlugin.py#L1348

PR #977, opened 2025-06-17
Replaced add folders with output directory button.

Silver0006:
I took the code from https://github.com/ciromattia/kcc/pull/969 and replaced the selectDir() function with it. Then replaced the Add Folders button and linked it to selectDir(). The button only works if the checkbox next to it is selected. The idea is you can set a output directory over multiple converts instead of re-selecting it with shift-click for each convert. Shift click for single converts still works

- fixes https://github.com/ciromattia/kcc/issues/745


Silver0006:
Should be good now

axu2:
Ok did some fixes, your checkbox wasn't saving because the names didn't match. (aka the box was called `enableDirectory` but you saved it as `enableDirectoryBox`)

Also fixed the 


> When I shift-click the convert button but choose "Cancel" in the output directory selection dialogue, the app starts the conversion process regardless (placing each output file in the directory that corresponds with its source file, as if I hadn't shift-clicked the "Convert" button in the first place.)

And fixed the confusing naming.

And made it so the folder is saved across sessions.

axu2:
Test builds: https://github.com/axu2/kcc/releases/tag/v7.5.1-custom-output-folder

axu2:
@Silver0006 What's your opinion of adding partially checked behavior, where it would automatically create a folder where the first file is?

Silver0006:
Are you saying a partial check would create a folder in the same directory as the first file and all output would be sent there? 

So the difference between all three options would be

- Unchecked exports to same directory as first file
- Partial creates a folder in same directory as first file and exports there
- Checked is custom directory

axu2:
@Silver0006 Yup exactly.

Though unchecked means existing behavior, next to source file, not first file.
For partial, I'd use os.mkdtemp with the name of the parent folder, maybe use pathlib.Path.parent
Checked is already implemented.

Silver0006:
I'll take a look at it when I get some free time.

axu2:
@Silver0006 Actually, I'll handle the partially checked behavior, it would be easier since I'm the one who wants this particular behavior and know the specific requirements.

Silver0006:
Ok

PR #978, opened 2025-06-17
Fixes to grayscale pages in forcecolor mode

9783e6:
Fixes bit depth/quantization for grayscale pages with forcepng + forcecolor options.
Now color pages will be saved in 8bit truecolor (24bit per pixel) and grayscale pages in 8bit (2bit/4bit/8bit dynamically based on given quantization palette after https://github.com/ciromattia/kcc/pull/976 is merged)

Closes https://github.com/ciromattia/kcc/issues/932

Slight optimization to JPEG grayscale pages with forcecolor option.
Consistently results in ~0.7% reduction in output file size.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -642,8 +642,8 @@ def imgFileProcessing(work):
             img.autocontrastImage()
             img.resizeImage()
             img.optimizeForDisplay(opt.reducerainbow)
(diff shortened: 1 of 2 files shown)


axu2:
I think your conditionals are too complicated and shouldn't call a separate function, it makes it harder to keep in my head.

I constructed tables of the final output, I think you can simplify the conditionals:

image

PR #981, opened 2025-06-18
Save images in GIF when output is set to MOBI and forcepng is used

9783e6:
Output stays the same as previously but processing time is cut in more than half.

Closes https://github.com/ciromattia/kcc/issues/533

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -325,8 +325,12 @@ class ComicPage:
     def save_with_codec(self, image, targetPath):
         if self.opt.forcepng:
             image.info["transparency"] = None
-            targetPath += '.png'
-            image.save(targetPath, 'PNG', optimize=1)
+            if self.opt.format == 'MOBI':
+                targetPath += '.gif'
+                image.save(targetPath, 'GIF', optimize=1, interlace=False)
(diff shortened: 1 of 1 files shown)


axu2:
Make sure this image works with GIF

- https://github.com/ciromattia/kcc/pull/494

I also left comments on all your other 2 PRs (3 total), please reply when you get a chance.

- #976
- #978

9783e6:
Yeah, it works to convert that png with transparency too

axu2:
Hmmm, I expected that image to fail, maybe something has changed in Pillow since then.

Anyways, I noticed that the generated epubs look like this:

```
    <item ... href="Images/kcc-0001-kcc.gif" media-type="image/jpeg"/>
```

That... might cause subtle issues, hmmm.

https://github.com/ciromattia/kcc/blob/2ecbf7d2e9636e64b0b9f2a794fd7f017a52cf35/kindlecomicconverter/comic2ebook.py#L337-L340

https://www.w3.org/TR/epub/#sec-core-media-types

axu2:
OK nope, even with that fixed png/gif on scribe is still completely busted.

Even tried setting the media type manually to jpeg and png, same result.

PR #986, opened 2025-06-21
only add display-block for kindle scribe

axu2:
- #972 

apparently this change broke the mupdf renderer for epubs, not sure how many people actually use that though. So I'll only add it when needed.

@9783e6


axu2:
I was playing around with removing stuff but this was the only way I could get it to work.

Issue #995, opened 2025-06-27
improved color detection


axu2:
some color images being labeled as bw by current algo

maybe just change the threshold?

PR #1003, opened 2025-06-30
Improved color detection

9783e6:
Improves the color detection algorithm to better handle images that contain only small regions of color but have a low overall color-to-pixel ratio. Previously, such images were often misclassified as grayscale, even when they did contain color. With this change, these pages are now correctly identified as color, reducing false negatives in color detection.

Additionally, this PR makes the color check run only when it’s actually needed, instead of running it up front every time. This avoids doing extra work if the color check isn’t used, which reduces unnecessary computation.

- Closes https://github.com/ciromattia/kcc/issues/995

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -647,7 +647,7 @@ def imgFileProcessing(work):
             img.autocontrastImage()
             img.resizeImage()
             img.optimizeForDisplay(opt.reducerainbow)
-            if opt.forcecolor and workImg.color:
+            if opt.forcecolor and img.color:
                 pass
             elif opt.forcepng:
                 img.quantizeImage()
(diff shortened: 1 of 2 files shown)


axu2:
Very clever! Looking at the wikipedia page of YCbCr makes it very clear why it's useful, you are basically ignoring the b&w portion of the image when doing the calculations and using color differences.

This can be even faster though.

Since the histograms are only 256 long, it's actually faster to use pure python.

```python
        cb_hist = cb.histogram()
        cr_hist = cr.histogram()

        cb_nonzero = [i for i, e in enumerate(cb_hist) if e]
        cr_nonzero = [i for i, e in enumerate(cr_hist) if e]
```

9783e6:
Yeah, using pure python in this case should actually be better, didn't think about that at first

PR #1027, opened 2025-07-12
Remove the rainbow effect on kaleido 3 by precisely attenuating the frequencies responsible instead of blurring the image

### LINKED ISSUE #771: Rainbow effect on grays on Kindle Colorsoft after converting with KCC
I've noticed that some grays have a rainbow effect on the Kindle Colorsoft. This doesn't ruin the experience, but it can be distracting.

I compared the same frame to a version from Amazon's official release and the rainbow effect on the Amazon version isn't as prominent. This issue happens with all compression settings (JPEG/PNG/mozJpeg)

Is there something wrong with my settings, or is this caused by something in the conversion process?

Here are my settings (6.2.1):

Screenshot 2024-11-04 at 9 28 18 PM

KCC version:
IMG_8841

Amazon version:
IMG_8843

Pictures taken with iPhone 16 Pro.

Thanks again.

Its-my-right:
Highly effective removal of rainbow effect artifacts on Kaleido 3 displays by significantly attenuating the responsible frequencies in the spectral domain (>= 0.3 cycles/pixel, oriented around 135°).
This method avoids blurring the entire image by precisely targeting problematic frequencies (at the cost of a ~ 18% processing time increase vs the blurring method on processed images).

Fixes #771

### DIFF
--- a/kindlecomicconverter/rainbow_artifacts_eraser.py
+++ b/kindlecomicconverter/rainbow_artifacts_eraser.py
@@ -63,12 +63,16 @@ def attenuate_diagonal_frequencies(fft_spectrum, freq_threshold=0.30, target_ang
     
     # Calculation of complementary angle
     target_angle_2 = (target_angle + 180) % 360
+
+    # Calulation of perpendicular angles (CFA is sometimes orientated at 135°, sometimes at 45°)
+    target_angle_3 = (target_angle + 90) % 360
+    target_angle_4 = (target_angle_3 + 180) % 360
     
     # Create angular conditions in a vectorized way
     angle_condition = np.zeros_like(angles_deg, dtype=bool)
     
     # Process both angles simultaneously
-    for angle in [target_angle, target_angle_2]:
+    for angle in [target_angle, target_angle_2, target_angle_3, target_angle_4]:
         min_angle = (angle - angle_tolerance) % 360
         max_angle = (angle + angle_tolerance) % 360
         
--- a/kindlecomicconverter/rainbow_artifacts_eraser.py
+++ b/kindlecomicconverter/rainbow_artifacts_eraser.py
@@ -14,8 +14,8 @@ def fourier_transform_image(img):
     
     return fft_result
      
-def attenuate_diagonal_frequencies(fft_spectrum, freq_threshold=0.3, target_angle=135, 
-                                     angle_tolerance=10, attenuation_factor=0.25):
+def attenuate_diagonal_frequencies(fft_spectrum, freq_threshold=0.30, target_angle=135, 
+                                     angle_tolerance=10, attenuation_factor=0.15):
     """
     Attenuates specific frequencies in the Fourier domain (optimized version for rfft2).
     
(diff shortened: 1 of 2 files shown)


Its-my-right:
If one does not know how to test this, I can provide outputs from kcc from source inputs provided by users.

I must admit I am not 100% certain the CFA on Kaleido 3 is always oriented at 135°, or if it depends on screen size or worst, varying among units.
On the 7,8" Kaleido 3 on my Inkpad Color 3, this is the case so the code attenuates high frequencies along this direction, but I do not have any other color e-readers so I could not check if this orientation is "universal" or not.

So it would be good if we could test this on other models / units

kindlecomicconverter/rainbow_artifacts_eraser.py
```diff
@@ -0,0 +1,127 @@
+import numpy as np
+from PIL import Image
+       
+def fourier_transform_image(img):
+    """
+    Memory-optimized version that modifies the array in place when possible.
+    """
+    # Convert with minimal copy
+    img_array = np.asarray(img, dtype=np.float32)
```
axu2, line 9: I'm assuming you need float for fourier transform, and that keeping it as an int (faster?) is not an option.

kindlecomicconverter/rainbow_artifacts_eraser.py
```diff
@@ -0,0 +1,127 @@
...
+    freq_y = np.fft.fftfreq(height, d=1.0)
+    freq_x = np.fft.rfftfreq(width_original, d=1.0)  # Use rfftfreq for the X dimension
+    
+    # Use broadcasting to create grids without meshgrid (more efficient)
+    freq_y_grid = freq_y.reshape(-1, 1)  # Column
+    freq_x_grid = freq_x.reshape(1, -1)   # Row
+    
+    # Calculate squared radial frequencies (avoid sqrt)
+    freq_radial_sq = freq_x_grid**2 + freq_y_grid**2
+    freq_threshold_sq = freq_threshold**2
```
axu2, line 48: This might be premature optimization, but is np.square faster by any significant margin?

axu2 (approved):

kindlecomicconverter/rainbow_artifacts_eraser.py
Its-my-right: Int is not enough for those types of calculations, and Numpy will convert it to float64 by itself with np.fft.rfft2() anyway. Since float32 need 2 times less memory and is fine fot image processing I thought it was a good compromise here

kindlecomicconverter/rainbow_artifacts_eraser.py
axu2: Either one is fine, but if there's a signficant speed increase may be worth looking into.

axu2: nevermind, it looks like it's the same https://stackoverflow.com/questions/25870923/how-to-square-or-raise-to-a-power-elementwise-a-2d-numpy-array

PR #1031, opened 2025-07-14
Add pdf support for Remarkable

Belgian-Coder:
Added support to export directly to PDF for Remarkable devices. These devices keep a lot of margin when using EPUB.


axu2:
@zamiam69 In case this interests you, are there margins on epub for you?

zamiam69:
Thanks. I will look into it and report back later this week in case I should encounter any problems.

Belgian-Coder:
Also unrelated to this pr, I ran into an issue with 1 comic because it had BMP images. When I looked at the history, support for this was removed in version 4.4 but I could not find why. Are we allowed to re-add it or is there another reason that it was removed?

axu2:
Found the commit where it was removed:

- https://github.com/ciromattia/kcc/pull/122/commits/84f58516a04a0fac369ea20e1ddb08b8bc97b33e

Probably a bug with an old Pillow version.

- https://github.com/python-pillow/Pillow/issues/1293

axu2:
Have you considered using MuPDF to create the PDF?

https://pymupdf.readthedocs.io/en/latest/recipes-images.html#how-to-make-one-pdf-of-all-your-pictures-or-files

MuPDF is going to be added soon as a dependency anyways in the other pending PR

Belgian-Coder:
I wasn't aware of that, I will take a look if I can switch it easily

axu2:
If you do, please close this PR and re-open a new one so it's easier to compare.

The flattenTree function might be useful.

axu2:
> Also unrelated to this pr, I ran into an issue with 1 comic because it had BMP images. When I looked at the history, support for this was removed in version 4.4 but I could not find why. Are we allowed to re-add it or is there another reason that it was removed?

Still looking forward to your bmp PR, it's a one line change @Belgian-Coder

PR #1076, opened 2025-09-08
experimental epub input with order

### LINKED ISSUE #277: EPUB Input [Enhancement]
It would be nice to be able to use EPUBs as an input for creating .mobi files. It should be relatively simple as they are used as an intermediary anyway.

axu2:
- fix #277 

@bust4cap @9783e6 @mwt @shihong5858

This is highly experimental, feel free to share your thoughts! This just simply copies all images in order according to html order. It's very hacky and I suspect it will fail on some inputs, but I'm not sure if those kinds of inputs ever happen

No metadata or anything yet.

https://github.com/axu2/kcc/releases/tag/v9.0.0-epub

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -545,7 +545,7 @@ class KCCGUI(KCC_ui.Ui_mainWindow):
             GUI.jobList.clear()
         if self.tar or self.sevenzip:
             fnames = QFileDialog.getOpenFileNames(MW, 'Select file', self.lastPath,
-                                                            'Comic (*.cbz *.cbr *.cb7 *.zip *.rar *.7z *.pdf);;All (*.*)')
+                                                            'Comic (*.cbz *.cbr *.cb7 *.zip *.rar *.7z *.epub *.pdf);;All (*.*)')
         else:
             fnames = QFileDialog.getOpenFileNames(MW, 'Select file', self.lastPath,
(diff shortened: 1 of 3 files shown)


bust4cap:
tried to convert several epubs to cbz.
my settings:
device: kindle paperwhite 11
right to left mode
custom gamma 1.0
color mode
disable processing

some notes:
-all images seem to get re-encoded even though disable processing is enabled.

-there are several files that didnt complete, all with the following error message: 
"no images detected, nested archives are not supported."

-some cbz arent in proper reading order but instead follow the image file name order of the epub

-if an epub had a blank page only as an xhtml instead of an image, no image is created so spreads are misaligned in those cases (didnt expect this to be implemented yet, just a suggestion for the future i guess)

is it going by spine order or just by html filename order?

axu2:
1) oh, I probably shouldn't be using pillow to save the image bytes, I should save the bytes directly. 

2) that's likely something broken in my html scanning code, do you have example html file where the image isn't being extracted?

3) I think I'm going by html order according to the opf , but maybe the library isn't doing what I expect.

bust4cap:
[test.epub.zip](https://github.com/user-attachments/files/22217132/test.epub.zip)
test epub that fails. i replaced all the images with blank ones due to copyright.

axu2:
Oof, yup, my code made a lot of assumptions that are broken by this file. Hmmmm

Issue #806, opened 2025-01-10
Skipped a lot of screens in Webtoon mode


highpepsi826:
**Describe the bug**
When using Webtoon mode, whatever options and output has been chosen (my output mainly with EPUB), there are many screens has been skipped pages missing (Eg. I have 001.jpg to 007.jpg, but the output has been stopped at middle of 006.jpg and no more output for 007.jpg)

**To Reproduce**
1. Choose "Webtoon mode", "Color mode", output with "Kobo Libra Colour" and "EPUB"

Below is the input original image: 
<img width="77" alt="image" src="https://github.com/user-attachments/assets/8c58c24d-c279-41b5-a4fa-6ca0482740fd" />

Here is the last output of the screen:
<img width="637" alt="image" src="https://github.com/user-attachments/assets/74f08892-c66a-4c86-ba86-d75896c22e15" />

As you can see, there is missing the end of the screen and no more output afterwards.

**Desktop (please complete the following information):**
 - OS: [Windows 11]
 - Device [Kobo Libra Colour]

axu2:
Unable to replicate with the single image provided. @highpepsi826 
![image](https://github.com/user-attachments/assets/c738abf7-3429-4746-94e6-d8b580df024a)

unzipping the epub shows that nothing was cut. 
![image](https://github.com/user-attachments/assets/998ddb0b-a589-42f4-a166-3a94f9c84aa4)

axu2:
> Hey I have a similar issue. When converting webtoon panels to CBZ format, it seems to add a lot of white or black at the bottom and top of the image. I have attached screenshots below

![Screenshot.2025-01-24.105154.png (view on web)](https://github.com/user-attachments/assets/813c61cb-8dc2-491f-95c5-a31a586265a4)
original image
[002.jpg (view on web)](https://github.com/user-attachments/assets/d3a5a441-4939-4c65-8792-fa22cd48f673)
first three slides

[kcc-0000-kcc.jpg (view on web)](https://github.com/user-attachments/assets/71a87138-0fee-4aed-9495-70c3544ad955)
[kcc-0001-kcc.jpg (view on web)](https://github.com/user-attachments/assets/947be69e-72cc-48ac-8ad1-ed286af7400c)
[kcc-0002-kcc.jpg (view on web)](https://github.com/user-attachments/assets/0eccc124-da9c-4eba-87c7-b1120f5588a5)

So whenever I would read this in a ereader on my phone, the comic is very cut off

[...]

Issue #607, opened 2023-10-22
splitting in webtoon


cristopher92:
I have an 11th generation kindle paperwhite, I would like to read a manhwa that has vertical view, however having very large pages like this image, but 
![immagine_2023-10-22_174917584](https://github.com/ciromattia/kcc/assets/95577226/c50a7cac-b0a1-44d0-94f7-9d69723fa0b3)
sometimes it puts all the pictures together, other times it cuts them off by itself. I noticed that if instead of kindle pw5 I put kindle 2 it cuts them fine, but is there any other way?

Issue #453, opened 2023-01-04
CLI Webtoon splitting is skipping key parts and dialog, when image backdrop is too dark


paolovinoya:
I usually try to prepare vertical comics for viewing through the Kindle. So it's only natural that I would pick the webtoon processing mode option
` -w, --webtoon       Webtoon processing mode`

I discovered that the output result of some of the chapters that I convert were not smartly cut.

The comic had several vertically long strips, 
<img width="215" alt="Exhibit A" src="https://user-images.githubusercontent.com/26787363/210482477-eb1af858-9c2a-4342-8f48-4a615108bc5b.png">

approximately 6 or 7 long images with multiple speech balloons.
<img width="330" alt="Exhibit B" src="https://user-images.githubusercontent.com/26787363/210483045-4c2f5336-9df2-4c1e-9df7-be09249dc0ab.png">

The end result was a mobi file with NO images within the black region and cutting back, only at the hourglass, at the end.

It managed to cut and display the lighter part of one strip with transition to black:
<img width="302" alt="image" src="https://user-images.githubusercontent.com/26787363/210483198-88e92d4e-1305-4643-9a8f-d6513cd08b25.png">

However the next page indicated that the process skipped out a lot of strips before - (skips a lot of black background) and ended up showing only the hourglass at the end.
<img width="293" alt="image" src="https://user-images.githubusercontent.com/26787363/210483342-131af001-ea30-4870-bfe8-70cfb42fd9ba.png">

Is there any option or feature where we can tweak the detection threshold of the images? Maybe disable or display the strip with a bit of overlapping at top or bottom parts of the cut image?

paolovinoya:
Updating this further, I managed to generate the debug output from kcc-c2p and the can see the detection mapped on where the white background is, but the black background comes undetected.

PR #1080, opened 2025-09-09
fixed width webtoon

axu2:
This chooses a fixed pixel width for the webtoon and doesn't do any panel detection at all, just scrolling a fixed length with some overlap. So gaps between panels can be large.

The code is hardcoded to 1072 width currently, so this will only work on 300 PPI devices.

Let me know if you have thoughts. 

Essentially does the same things as this app: https://www.youtube.com/watch?v=GYH9tKC5F6M

https://github.com/axu2/kcc/releases/tag/v9.0.0-webtoon-fixed-width

- #806
- #607 
- #453

@highpepsi826
@shor-tee
@cristopher92
@paolovinoya

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1520,8 +1520,8 @@ def makeBook(source, qtgui=None):
     cover = image.Cover(cover_path, options)
 
     if options.webtoon:
-        y = image.ProfileData.Profiles[options.profile][1][1]
-        comic2panel.main(['-y ' + str(y), '-i', '-m', path], qtgui)
+        x, y = image.ProfileData.Profiles[options.profile][1]
+        comic2panel.main(['-y ' + str(y), '-x' + str(x), '-i', '-m', path], qtgui)
     if options.noprocessing:
         print("Do not process image, ignore any profile or processing option")
     else:
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1402,6 +1402,8 @@ def checkOptions(options):
         options.righttoleft = False
         options.upscale = True
         options.hq = False
+        options.white_borders = True
+        options.bordersColor = 'white'
     # Disable all Kindle features for other e-readers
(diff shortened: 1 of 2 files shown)


paolovinoya:
I confirmed that it solved my report on the issue where the original routine appears to not be able to detect transitional background. The supposed missing pages of my test chapter is showing up with the correct sequence - no dialogue was dropped.
image
While it's a shame that it's hardcoded to 1072, it practically solved my problem statement in #453 though I wish there would be more flexibility in the future by configuring a width variable, both on UI and on command-line.

axu2:
@paolovinoya what is your device resolution and what fixed width would you choose? I want to know how people would use this feature

paolovinoya:
I'm currently using a Kindle 10th Gen (600x800 167ppi) and a Kindle Paperwhite Signature 11th Gen (1072×1448 300ppi). I use the KV choice in the model types to choose in the command line to fit both my devices. I'm not complaining at this point. The resulting image may be too tall, and the resulting effect was that my eyes would still scan from top going down, in the same way as a vertical comic should. The resolution appears to fit the Paperwhite, though I am yet to try this out on my smaller and lighter Kindle. 

My use case is to make my Kindle be my travel companion while taking my train commute - which is 2 hours one way. I tried using my phone as a reading source, but leaves me with a low battery status at the end, or distracting phone notifications here and there. A tablet was a bit too heavy (Samsung Galaxy S7FE), and went through near-misses by accidentally dropping it in my daily commute.

Not sure if there's any difference to mention, but I often use the container release, so I automate this with a script to batch convert my collection of vertical comics.

axu2:
@paolovinoya That's not what I meant regarding fixed width.

For example, I'm using a 1860x2480 Kindle. Fixed width of 1072 means the output image is 1072x2480, which takes up ~60% of the screen's width.

With your 1072x1448 paperwhite, it would output 100% of the screen width.

Approximately how much of the screen's width is your preferance? I'll probably make it adjustable but curious to know your ballpark of how you would use it.

I need to fix the large gaps betwen panels.

PR #838, opened 2025-02-28
Optimizations

utopiafallen:
Sorry, not sure how to cherry pick a specific commit for the pull request so here's everything from my fork :)

PR #1089, opened 2025-09-19
Fix [WinError 32] The process cannot access the file because it is being used by another process

axu2:
Multiframe gif images are not closed automatically by Pillow. So we make a copy and close the file manually. This doesn't seem to affect speed.

https://pillow.readthedocs.io/en/stable/reference/open_files.html#image-lifecycle

test builds:
https://github.com/axu2/kcc/releases/tag/v9.1.0-gif

@utopiafallen I think this sounds similar to an issue you had before? The error occured when it tried to delete

- https://github.com/ciromattia/kcc/pull/838/commits/35f20e82230a27f9f01eedb51043c31c3858cb27

### DIFF
--- a/kindlecomicconverter/image.py
+++ b/kindlecomicconverter/image.py
@@ -151,7 +151,8 @@ class ComicPageParser:
         # Detect corruption in source image, let caller catch any exceptions triggered.
         srcImgPath = os.path.join(source[0], source[1])
         Image.open(srcImgPath).verify()
-        self.image = Image.open(srcImgPath)
+        with Image.open(srcImgPath) as im:
+            self.image = im.copy()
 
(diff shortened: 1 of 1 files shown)


axu2:
@Lukas-Holecy maybe this fixes your past memory related issue?

axu2:
@utopiafallen Was this the problem you had before?

utopiafallen:
Not specifically with gifs since I wasn’t converting any gifs, but yea, sounds like the same issue and why I threw in the retry loop that you’ve brought in from my initial PR.

axu2:
@utopiafallen OK, so maybe it wasn't this specific related gif issue, but this should fix your issue anyways, since now KCC will immediately close the file handle after copying it to memory.

PR #1092, opened 2025-09-29
feat: update image.py to support remarkable paper pro move profile

luismelodev:
Given that the reMarkable Paper Pro Move was launched a few weeks ago, I'm creating a new profile to support the device.


axu2:
Nice! If you prefer not to have a colored border just partially check the margin color box.

Issue #522, opened 2023-05-15
Output of Volume metadata


jaroslawjanas:
So right now we have something like this
```python
if xml.data['Volume']:
    titleSuffix += ' V' + xml.data['Volume'].zfill(2)
```
Which produces the following title
`Title V##`
personally, I am more of a
`Title Vol.##`
kind of guy, it's what I am used to and what seem to be most popular.

As such I'd like to either
**Option A**
```python
if xml.data['Volume']:
    titleSuffix += ' Vol. ' + xml.data['Volume'].zfill(2)
```

**Option B**
```python
if xml.data['Volume'] && options.forV:
    titleSuffix += ' V' + xml.data['Volume'].zfill(2)
elif xml.data['Volume'] && options.forVol
    titleSuffix += ' Vol. ' + xml.data['Volume'].zfill(2)
```
or something along those lines.

And since we are on the topic it might be worthwhile to implement global app settings UI.
It would be useful for user-specific options and so on, things like language preferences #501, and naming conventions like the one above.

I don't know what the exact vision for the UI is, so I'd love to hear what you think about it.

Nizzuta:
I support this, I went crazy trying to change the metadata on calibre, had to download the source and change it myself. It's easy for someone with at least a little technical knowledge, but most people won't do that. A GUI option, or a least a text file config would be great

jaroslawjanas:
The GUI would be probably the biggest hassle.
This also puts to question the current UI layout.

But I think **global** settings such as the one above should have their own _Settings_ or _Preferences_ GUI.

I can maybe give this a shot at a much later date since my next two months are somewhat booked.

axu2:
@vinhtq115 this is what you want?

vinhtq115:
Yeah it's kinda like this.

axu2:
I got to be honest, might be easier to just store metadata in a special file next to the original cbz/cbr. There's a lot of difficulties storing it directly in the original file, like if we have rar file that can't be modified on lots of operating systems.

vinhtq115:
Probably per file .xml that matches the cbz filename is better.

jaroslawjanas:
> I got to be honest, might be easier to just store metadata in a special file next to the original cbz/cbr. There's a lot of difficulties storing it directly in the original file, like if we have rar file that can't be modified on lots of operating systems.
> 
> Maybe per file .ini or .xml?

I am lost here. isn't that like... a lot of work for the user?

axu2:
[...]

PR #1096, opened 2025-10-04
Add Title field to Metadata editor

kiryl85:
As PR title describes, `Title` field will be added to Metadata Editor

obraz


bust4cap:
the label is missing a ":"

kiryl85:
@bust4cap 
indeed, please check now

axu2:
LGTM! Good idea to pre-populate the title field as the filename.

kiryl85:
> LGTM! Good idea to pre-populate the title field as the filename.

I'm working on that already :)

axu2:
@kiryl85 wait but it does it now? After I fixed some older bugs in above pr

- #1099

kiryl85:
@axu2 from what I see mobi title (the way how it is displayed on kindle) is either: file name, custom title (but when run from command line), metadata based title (issue number + volume number + series) or metadata title (if checked in gui).
I was thinking on changing metadata title checkbox behaviour in the way, that metadata title is either included in existing behaviour (=+) or it overwrites the title completely (the way how it is currently)

axu2:
> that metadata title is either included in existing behaviour (=+)

Oh I see, another use case. Sounds useful. So like 

`<Series> VOL <NUMBER>: <TITLE>`

Maybe 

- #522 

is relevant.

kiryl85:
Yes, let me create PR today and you will see. That should be enough to close that issues.

PR #1100, opened 2025-10-06
fix chapternames overwrite with ComicInfo.xml bookmarks

kiryl85:
Currently if there are bookmarks defined in ComicInfo.xml file, chapternames can be overwritten only when comic structure is flat, which means all chapters/images are placed in one directory.

With this change tree structure is completely ignored. This should fix issue mentioned in: https://github.com/ciromattia/kcc/discussions/1097


axu2:
LGTM, though in the future I would prefer 1 PR per idea. The vol thing is separate idea than chapters.

kiryl85:
indeed, sorry for that. `Vol.` was intended to be in a different PR

PR #1101, opened 2025-10-07
Update title generation

kiryl85:
Title can now be:
- generated as is (default method),
- generated as is (default method) with addition of metadata Title,
- based on metadata Title only, and in addition
- overwritten with custom value without editing metadata, Field added in GUI

Entry sample
metadata_sample

Test cases:

1. Default title
default_title
default_title_result

2. Default title with metadata title
default_title_with_metadata_title
default_title_with_metadata_title_result

3. Metadata title only
metadata_title_only
metadata_title_only_result

4. Custom title
custom_title
obraz

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -371,6 +371,11 @@ class WorkerThread(QThread):
             except Exception as e:
                 print('Fusion Failed. ' + str(e))
                 MW.addMessage.emit('Fusion Failed. ' + str(e), 'error', True)
+        elif len(currentJobs) > 1 and options.title != 'defaulttitle':
+            currentJobs.clear()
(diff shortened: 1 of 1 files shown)


axu2:
A default title box doesn't really make sense to me since KCC does bulk conversions. 

Is that feature purely meant for when converting a single item?

kiryl85:
Yes , that was the idea. It could be extended to file fushion as well, so either one element on the list, or many elements with fushion.

If still not ok, I'm ok to drop title GUI related commit

axu2:
Oh... for file fusion, yes that's a nice feature combo. Let's keep that. @Silver0006

kiryl85:
@axu2, I have added fusion case

1. Custom title with many sources without fusion
custom_title_many_source_fusion_off
custom_title_many_source_fusion_off_result


2. Custom title with many sources with fusion
custom_title_many_source_fusion_on
custom_title_many_source_fusion_on_result

PR #1102, opened 2025-10-10
Open metadata editor after doubleclick on source

kiryl85:
Open source in metadata editor after double click. Works for both, cbz file and comic directory

I think this one will be pretty handy :)


axu2:
Yea, that is useful!

axu2:
@kiryl85 Feel free to search for open metadata related issues and see if any catch your eye!

PR #1108, opened 2025-10-13
Keep sorted and unique list of sources

kiryl85:
- sort items on the sources list alphabetically -
- don't add item if it is already on the list


kiryl85:
@axu2 I've rebased latest master over here

axu2:
I guess this is fine, I was initially worried about adding an N^2 and NlgN operation here but n doesn't get high enough to warrant worrying realistically.

I can see this being especially useful with the progress bar feature when you have a lot of sources from a lot of places.

Issue #364, opened 2020-06-17
Feature request - progress bar on bulk jobs


wardrich:
Maybe I'm just dumb and overlooked it, but if it's not there it would be great to get a progress bar for bulk jobs to see how far along it is.  Even if it only increments each time a book in the list is done.

This program is otherwise really awesome!

axu2:
@abraxas86 This is nearly ready

PR #1109, opened 2025-10-13
Update progress notification for bulk jobs

kiryl85:
- fix  https://github.com/ciromattia/kcc/issues/364


axu2:
Sorry for the delay, please double check the additional changes I made. @kiryl85 

Then I'll merge.

Is there any particular reason you chose to index each loop iteration instead of a simpler enumerate?

kiryl85:
yeah, you are right. no particular reason, I simply overlooked that possibility. Indeed, enumerate makes more sense here.

Thanks for pointing that out.

axu2:
@kiryl85 Nice feature for when running large sources. Thanks for the contribution.

PR #1155, opened 2025-11-08
Optimized Docker image

### LINKED ISSUE #1154: Possibility of optimizing docker image
I'm curious to know if there is any possibility of optimizing docker image, as it is way too big (1.06gb) due to installing all dependencies in requirements.txt, when PySide6 and Raven are only used on the GUI, which is (as I understand based on the documentation) not intended to be used in the Docker version of the app.
I have no problem implementing the changes, but I want to know if there is any reason for the image containing the whole app (and using two dockerfiles instead of one)
Thanks

joseenrique61:
Closes #1154 
### Refactor Docker build process for simplicity and optimization

This pull request overhauls the Docker build process to follow modern best practices, resulting in a simpler, faster, and more secure build that produces smaller final images.
Summary of Changes:

Unified Dockerfile:

- The separate Dockerfile-base has been eliminated and its logic merged into a single, multi-stage Dockerfile.
- This removes the need for a two-step build process and a separate docker-base-publish.yml GitHub workflow, which has been deleted. The entire image is now built atomically.

Optimized Multi-Stage Build:

- The new Dockerfile uses a proper multi-stage pattern with a builder stage and a final stage.
- The builder stage installs all build-time dependencies (build-essential, -dev packages, etc.).
- The final stage starts from a clean base image and only copies necessary runtime dependencies and application code from the builder, significantly reducing the final image size and attack surface.

Flexible Entrypoint Script:

- A new entrypoint.sh script has been introduced to act as the container's entrypoint.
- This allows for greater flexibility, enabling different modes of operation (running the c2e or c2p) controlled by the KCC_MODE environment variable.

Dependency Optimization:

- A new requirements-docker.txt has been created.
- PySide6 and raven have been removed from this list as they are GUI-specific dependencies and are not required for the headless CLI environment in the Docker container.
- The .dockerignore file has been updated to exclude unnecessary files, ensuring a clean and efficient build context.

Build Process Improvements:

- The build logic now leverages Docker's layer caching more effectively by copying requirements-docker.txt and installing dependencies before copying the application source code.
- Removed apt-get upgrade in favor of relying on the base image maintainers for security patches, ensuring build reproducibility.

Testing:

The new Docker image has been successfully built and tested (superficially, as there are no automated tests available) on the following architectures:

- linux/amd64: Works as expected.
- linux/arm64: Works as expected.

Testing for linux/arm/v7 was attempted, but the build process under QEMU emulation was prohibitively slow due to the heavy compilation required for some Python packages on this architecture.

### DIFF
--- a/.github/workflows/docker-base-publish.yml
+++ b/.github/workflows/docker-base-publish.yml
@@ -3,8 +3,8 @@ name: Build and Publish Base Docker Image
 on:
   workflow_dispatch:
   push:
-    tags: 
-      - 'v*.*.*'
+    branches:
+      - master
 
     paths:
       - 'requirements-docker.txt'
@@ -27,9 +27,10 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
-      - name: Extract tag name
-        id: tag_name
-        run: echo "TAG=${{ github.ref_name }}" >> $GITHUB_ENV
+      - name: Set Release Date
+        id: release_date
+        run: |
+          echo "release_date=$(date --rfc-3339=date)" >> $GITHUB_OUTPUT
 
       - name: Build and push
         uses: docker/build-push-action@v6
@@ -38,39 +39,7 @@ jobs:
           file: Dockerfile-base
           push: true
           tags: |
-            ghcr.io/${{ github.repository_owner }}/kcc:base-${{ steps.tag_name.outputs.TAG }}
             ghcr.io/${{ github.repository_owner }}/kcc:base-latest
-          cache-from: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache
-          cache-to: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache,mode=max
-
-  build_and_publish_image:
-    needs: build_and_publish_base_image
-    runs-on: ubuntu-latest
-    steps:
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
-
-      - name: Extract tag name
-        id: tag_name
-        run: echo "TAG=${{ github.ref_name }}" >> $GITHUB_ENV
-
-      - name: Build and push
-        uses: docker/build-push-action@v6
-        with:
-          platforms: linux/amd64,linux/arm64,linux/arm/v7
-          push: true
-          tags: |
-            ghcr.io/${{ github.repository_owner }}/kcc:${{ steps.tag_name.outputs.TAG }}
-            ghcr.io/${{ github.repository_owner }}/kcc:latest
+            ghcr.io/${{ github.repository_owner }}/kcc:base-${{ steps.release_date.outputs.release_date }}
           cache-from: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache
           cache-to: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache,mode=max
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -35,17 +35,17 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
-      - name: Extract tag name
-        id: tag_name
-        run: echo "TAG=${{ github.ref_name }}" >> $GITHUB_ENV
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository_owner }}/kcc
 
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
           platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: true
-          tags: |
-            ghcr.io/${{ github.repository_owner }}/kcc:${{ steps.tag_name.outputs.TAG }}
-            ghcr.io/${{ github.repository_owner }}/kcc:latest
-          cache-from: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache
-          cache-to: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:base-cache,mode=max
+          tags: ${{ steps.meta.outputs.tags }}
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:final-cache
+          cache-to: type=registry,ref=ghcr.io/${{ github.repository_owner }}/kcc:final-cache,mode=max


axu2:
> The build logic now leverages Docker's layer caching more effectively by copying requirements-docker.txt and installing dependencies before copying the application source code.

Excellent, I believe this is why we had two dockerfiles before, the base image was only built when requirements.txt was changed. This takes 2 hours on armv7 on Github actions.

BTW, how much smaller are the images? Before and after?

joseenrique61:
Yes, I think it could make sense to have a base image if we had more images that inherited from it, but we only have one, and it only copied the source code to the container, making it unnecessary.
Respecting the size of the image, it was ~1.06GB before, it is ~400MB now, so it is more than a 50% decrease in size.
Also, as I said, I tried building for arm/v7, but after 40+ minutes I gave up

joseenrique61:
By the way, I saw there are some critical errors on python-3.13-slim-bullseye. I didn't update it to a newer version of Debian because it is a major change that could require more tests, but it wouldn't be a bad idea thinking from a security perspective

axu2:
Since the 2 hour build on armv7 is so long, maybe it would be a good idea to "fix" the versions of numpy and pymupdf to an explicit version, since I believe those are the dependencies that take a long time to compile from source. So maybe pip install those as a separate initial step in the dockerfile, then install the rest of the `requirements-docker.txt`

That way, development would be a lot faster since it would cache the numpy/pymupdf compiles. (Note, it's been a while since I've used docker so I might misremeber how docker works.)

joseenrique61:
I think it does work that way, but cache has to be configured in github actions if I remember well

axu2:
In case it was unclear, the old base image wasn't built automatically every release, it was manually rebuilt when requirements.txt changed. So the 2 hour process didn't happen every release.

axu2:
OK, github caches will not work since time between releases can be more than 7 days:

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB. Once a repository has reached its maximum cache storage, the cache eviction policy will create space by deleting the caches in order of last access date, from oldest to most recent.

https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy

So that's why we had multiple dockerfiles, since we can't cache.

axu2:
Looks like GitHub limit is 6 hours per job, ~30 hours per month total.
https://docs.github.com/en/actions/reference/limits

But 2 hour build time is annoying for development. And a waste of compute time. Especially when doing test and beta builds.

How easy is it to restore the base dockerfile? I believe it was two dockerfiles since the 1st one was acting as a cache

And feel free to try 3.11 or 3.12 if you need.

joseenrique61:
Ooooh I get now, that makes sense. I will create a base image then, keeping the optimizations made in this one.
And in fact, it would be a good idea to have the versions of packages fixed, mainly for reproducibility, since changes in one package could break the app without us knowing as we don't have automated tests.

axu2:
A separate pr might be easier to compare

joseenrique61:
Ok. I divided the image into two, keeping the same optimizations as before. Also, changed the workflows to use docker/build-push-action instead of the SDR enthusiasts they had before, as this is an official action and allows to put tags on images, so we can use base-latest in the final image instead of base-xxxx.

Also automated the build process of both images. The build is still done when a version is released (tag v*.*.*). The workflow now consists of two cases:
1. Files requirements-docker.txt or Dockerfile-base are updated: both the base and the final image are rebuilt.
2. Any other file than those (or documentation files) are updated: only the final image is rebuilt.

joseenrique61:
I noticed there are some packages missing in the dockerfile. I'm adding them now.

.github/workflows/docker-base-publish.yml
```diff
@@ -1,34 +1,76 @@
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Extract tag name
+        id: tag_name
+        run: echo "TAG=${{ github.ref_name }}" >> $GITHUB_ENV
```
axu2: same here

.github/workflows/docker-publish.yml
```diff
@@ -15,19 +15,37 @@ on:
...
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Extract tag name
+        id: tag_name
+        run: echo "TAG=${{ github.ref_name }}" >> $GITHUB_ENV
```
axu2: here too

joseenrique61:
Ok, I had to move some things around. First, I already tested all necessary packages are installed during the build process of the base image.
Second, the base image was activating whenever a tag was published, so I changed it to only run when a push to the master branch was made, and only on the files mentioned in my previous comment. The final image will still activate on a tag push.
Because of this change, the git tag value couldn't be used anymore for the base image tag (as it is not activated by a tag push), so I changed the image to have the tags "base-latest" and "base-date", with date in YYYY-MM-DD format.
I also changed the docker-publish.yml workflow to use docker/metadata-action as suggested in https://github.com/docker/metadata-action?tab=readme-ov-file#about, for better tagging of the final image.

axu2:
LGTM. I'm gonna run the build in prod and make sure everything works.

Out of curiosity, did you use any LLM assistance? 

>Second, the base image was activating whenever a tag was published, so I changed it to only run when a push to the master branch was made, and only on the files mentioned in my previous comment.

sounds even better! Only builds when a push to master changes the dockerfile-base or requirements-docker.

Now just to wait for 2 hour build to finish, too bad it's not cached across forks but that's fine.

joseenrique61:
I did use Gemini 2.5 Pro for the first refactor and to explain me how multi-architecture images work, as I don't have that much knowledge in building for docker. It's a really great tool for that sort of things.

And yeah, it took 2 hours and 8 minutes in my fork to build, but after that any subsequent build was just 1 min. It would be great if the cache in github could last longer than 7 days.

Let me know how it goes.

axu2:
Wait, does this still work with kindlegen? A user just needs to add it manually. What's against the license is redistributing the binary directly

joseenrique61:
It should. The user just has to add the -v "path-to-kindlegen:/opt/kcc" to the command of docker run.

axu2:
> It should. The user just has to add the -v "path-to-kindlegen:/opt/kcc" to the command of docker run.

OK, I think I see a reference to that in the wiki.

Also, any particular reason why license was removed from LABEL? Maybe other labels are important.

joseenrique61:
There were some labels duplicated, I think I accidentally deleted that one. I'll add it back after you test the image and check it's working

axu2:
@joseenrique61 built, can you confirm if everything works, including kindlegen?

https://github.com/ciromattia/kcc/pkgs/container/kcc

joseenrique61:
I actually tried downloading kindlegen, but never found where to get it from

axu2:
@joseenrique61 Made a test 9.3.0 release

joseenrique61:
But where do you get it from?

axu2:
I'm not sure, I'm not a linux user. Are you using linux?

joseenrique61:
Yeah, I use Fedora as my main OS. In the wiki it says the name of the file you have to download for the flatpak version, but not where to get it from. 
I actually have never used kindlegen, as I have my kindle jailbroken and only use .cbz

axu2:
Amazon used to offer a direct download of the linux version of kindlegen, I'm sure there's an archive of it.

axu2:
> But doesn't kindle support epub natively or something? Or is that only when you send it to the device's email and amazon converts it?

kindlegen converts an epub to mobi/azw3 (kindle native format)

kindlegen runs in the amazon cloud when you send epubs and cloud delivers azw3

axu2:
Also, final shouldn't be cached, only base. And we don't need the `base-cache`, right? we already have `base-date` and `base-latest`

image

axu2:
Wait, this runs every time I push to master? Maybe this should only be manually run.

image

joseenrique61:
You're right, for some reason the workflow executed whenever you pushed to master, even though in the [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore) it says the following:

> If you define both branches/branches-ignore and paths/paths-ignore, the workflow will only run when both filters are satisfied.

And regarding the cache, the final-cache can be removed, as it is not that long of a build; but doing some research I found that when you pass the parameters "cache-to" and "cache-from" to the docker/build-push-action, it can store the cache as a [special image](https://docs.docker.com/build/ci/github-actions/cache/#registry-cache) in the registry. I'm not quite sure, but I think that means the cache doesn't have an expiry date, as it is just stored as another image.

I also solved the problem with the workflow executing every time you push to master, and tested it on my fork. Do I open another PR for it?

axu2:
Yea another PR

Wait does registry cache mean you can use single docker file?

PR #1157, opened 2025-11-10
Optimize docker image

joseenrique61:
Continuation of #1155 
### Changes made:
- Unified the workflows for the docker images
- Both images will be built only on tag pushes on files that are not documentation
- The base image will only be built when requirements-docker.txt or Dockerfile-base are changed
- The final image will always be built, but will wait for the base image to be built in case it needs to. In case it doesn't (skipped because the files mentioned above were not modified) the final image will still be built.
- Re-introduced LABELs in the images

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,5 @@ packaging>=23.2
 mozjpeg-lossless-optimization>=1.2.0
 natsort>=8.4.0
 distro>=1.8.0
-numpy==2.3.4
-PyMuPDF==1.26.6
+numpy>=1.22.4
+PyMuPDF>=1.18.0


axu2:
I'll review soon, but initial thoughts:

1) Maybe the 2 hour build should never be automatically run, it should always be manually run. since requirements change so infrequently. I would rather not this be run automatically on forks. In fact, maybe this was why it ran on master for me, since it was bundled with dockerfile change when I pulled from upstream. Or maybe just add a check to only run automatically on main/your repo.
2) 3rd party actions always make me nervous, especially since the specific action you used was compromised in the last 8 months.

axu2:
Maybe read from the cache of the main repo ciromattia.

axu2:
Why did you use v41? of tj-actions? 

image

joseenrique61:
Sorry, I confused the 7 for a 1 in v47 (the current version)
In any case, I think you are right, maybe it's better to manually build the base image as it rarely changes. That would also solve the problem of having an external action in the workflow.

axu2:
I feel comfortable pinning the hash. 
I think pulling from the ciromattia cache also fixes my other problems. Take a look?

And can you confirm kindlegen still works?

Overall, I think this is nearly ready. 

sorry for making the mistake about multiple dockerfiles and github caches, with our correct info, is it still possible to combine the dockerfiles?

Wait since we are caching from ciromattia we don't even need to check for files changed?

.github/workflows/docker-publish.yml
```diff
@@ -1,12 +1,12 @@
...
 on:
   workflow_dispatch:
   push:
-    # Publish semver tags as releases.
-    tags: [ 'v*.*.*' ]
+    tags: 
+      - 'v*.*.*'
     
-    # Don't trigger if it's just a documentation update
+    # Don't trigger if it's just a documentation update or a base image update
```
axu2: This comment is no longer relevant?

Dockerfile
```diff
@@ -1,19 +1,25 @@
...
-LABEL org.opencontainers.image.description='Kindle Comic Converter'
-LABEL org.opencontainers.image.documentation='https://github.com/ciromattia/kcc'
-LABEL org.opencontainers.image.source='https://github.com/ciromattia/kcc'
-LABEL org.opencontainers.image.authors='darodi'
-LABEL org.opencontainers.image.url='https://github.com/ciromattia/kcc'
-LABEL org.opencontainers.image.documentation='https://github.com/ciromattia/kcc'
-LABEL org.opencontainers.image.vendor='ciromattia'
-LABEL org.opencontainers.image.licenses='ISC'
-LABEL org.opencontainers.image.title="Kindle Comic Converter"
+FROM ghcr.io/ciromattia/kcc:base-latest
```
axu2: With our correct info, can we go back to original combined dockerfile?

axu2:
I think those above commits should do it, can you double check?

SDR was nice, but it doesn't seem to be compatible with registry cache

and an initial pip install of numpy and pymupdf would be nice, since they take forever to compile.

joseenrique61:
Yes, with the cache I think we wouldn't need two images

Everything seems fine, just needs to adjust the names of the tags to not use "base", as there is no base image anymore

And you're right, I'm gonna see what versions of numpy and pymupdf are stable and being used in the project now and install them in a previous step on the Dockerfile

axu2:
Cool, I'll let you handle the names

Pymupdf takes longer to build. So maybe it should be first?

I don't think either is likely to change. 

Eh I think pymupdf is more likely to change actually, make it 2nd. It might come with an awesome new feature in future. Whereas numpy is extremely stable.

joseenrique61:
Ok. I'm testing it now on my fork. We have to wait for the two hour build as the layers changed and I changed the name of the cache.

joseenrique61:
Ok, done. It runs good on my fork, and adds the tags:
image

I moved the tags to the docker/metadata-action. They are still fixed: latest and release date will always be generated. The only tag that will vary will be the version one, as it only activates when the workflow is triggered because of a tag being pushed. If the workflow is triggered manually, it will not generate that tag.

I also divided the installation of packages into three steps:
1. Install numpy
2. Install PyMuPDF
3. Install the rest of the packages

I put numpy and PyMuPDF a fixed version to prevent the layer from being invalidated.
Also, I removed the package "unrar-free" from the final stage of the image, as it is only used for Fedora.

axu2:
Your changes are on a different branch, add them to this branch or open a new PR. @joseenrique61

joseenrique61:
These are the commits:
imagen

I made the pushes some hours ago

axu2:
Ah, guess I missed that. 
image

axu2:
OK, fixed the order, the way you had it if anything was added to requirements.txt, it would cause numpy and pymupdf to be rebuilt. This way fixes it.

joseenrique61:
Oh yeah I missed that. Guess it's done now.
Respecting the tests, I really couldn't find the kindlegen executable for linux, so I didn't test that

axu2:
Alright, this is ready for wider testing. This will hopefully be the FINAL 2 hour build.

axu2:
@darodi In case you want to weigh in since you initially wrote the docker

joseenrique61:
In fact, we needed that line of ln -s for kindlegen to work. I'm opening another PR for that.

PR #1176, opened 2025-11-30
Sync setup install_requires with requirements.txt

kiryl85:
I've noticed that `setup.py` uses slightly different packages versions than `requirements.txt` and it couldn't stop bothering me.
So here is the PR to fix that.

Also PyMuPDF was listed twice with different versions.
Also pyinstaller, even though it is needed to build binaries, was missing in `requirements.txt`.

Feel free to close that PR if mentioned differences doesn't matter.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,4 +10,3 @@ natsort>=8.4.0
 distro>=1.8.0
 numpy>=1.22.4
 PyMuPDF>=1.18.0
-pyinstaller


axu2:
I guess this is fine, some notes:

pyinstaller is not a runtime dependency. I don't want to install it in some circumstances.
The lowest common denominator is `requirements-win7.txt`. Any slight variations don't matter.
pyside6 6.10 drops macOS 12 support, I'm not ready to do that but that shouldn't stop anyone else. I might do it if Python 3.14 is a lot faster, since only 6.10 supports 3.14. And move macOS 12 to the legacy build

PR #1177, opened 2025-11-30
set tabulation order for metadata editor fields

kiryl85:
- loop through fields in from top to down order

**Note**: this PR might raise some conflict once https://github.com/ciromattia/kcc/pull/1178 is merged, which will be fixed afterwards


axu2:
Instead of manually specifying the tab order after the fact, wouldn't it be cleaner to just change the order of the original fields? The title field is last, I guess I missed that when it was first added.

kiryl85:
thanks @axu2
actually, editing UI is a little bit tricky, whenever something new is added via GUI editor and saved, the order of elements is completely mixed :/ so now I understand why earlier after generating new ui files there were so many changes, even thou I was adding only one field in specific place.

By specifying manual tabs order we can ignore all other changes that are made automatically in *.ui files (which happens after saving the changes we have made) and in generated py files (after runing `gen_ui_files` script) and be sure that tabs order remains correct.

Other option (less friendly but much better and easier for code reviews) is to make sure that newly added fields are placed in correct order before generating new ui files. Then indeed only our changes would be highlighted by diff and tab order wouldn't be needed at all.

I kind of lean to the second option and would like to apply the same approach to KCC ui file.

axu2:
Just checking, you used this feature to make the tab order right? 
image

I guess it's probably fine since it's easy to change. I would probably add something to readme about this for devs

kiryl85:
Sorry, I didn't have time to respond :)
but yes, I've used this functionality

PR #1178, opened 2025-11-30
set tabulation order for KCC fields

kiryl85:
- loop through fields in more organized order including fields which visibility depends on some checkboxes state

**Note**: this PR might raise some conflict once https://github.com/ciromattia/kcc/pull/1177 is merged, which will be fixed afterwards


axu2:
Manually specifying in this case makes sense.

axu2:
OK, the rc changes were all removed, just an artifact of the particular machine you built on.

Honestly maybe gen_ui_files shouldn't include it unless rc actually changed...

Issue #1163, opened 2025-11-17
Kindle Scribe blank pages


jaroslawjanas:
**Describe the bug**
The bottom 1/4 of a page is often cut. Not always but often enough to be an issue.

![Image](https://github.com/user-attachments/assets/1130cf12-372b-4467-b89b-4f87e819d574)

On top of that I have been getting blank pages here and there.

**Settings**

<img width="490" height="642" alt="Image" src="https://github.com/user-attachments/assets/c17dc73e-71d8-4547-9715-7957b291965f" />

**Desktop:**
 - OS: Windows 10
 - Device: Kindle Scribe 1st gen

**The Issue**
I believe the issues arises as an effect of the `Inter-panel crop`.
I am trying to eliminate vertical lines in double pages and it works to some extent (they get narrower) but in return I get this issue.

Now, I haven't narrowed it down to that specific setting but I will update the issue when I do.

**Additional Notes**
How come I can't select just `vertical lines` in the `Inter-panel crop`.

axu2:
That's unusual.

I would uncheck upscale, uncheck panel view. In fact, why are you using 2 panel? Scribe is big enough that you don't need to zoom.

Usually those problems only occur for me if I turn pages unreasonably fast.

That problem never occurs for me when I'm just reading normally.

I really doubt the crop options affect anything.

Maybe it's cause I use an old version of kindlegen from Kindle Comic Creator?

jaroslawjanas:
I think I figured it out, it seems to be the `Cropping mode`.

<img width="490" height="642" alt="Image" src="https://github.com/user-attachments/assets/daaf0250-9e92-4120-a8b8-83653f9b6943" />

Here is the funny thing. I switched the output to **CBZ** to investigate the output directly rather than through my kindle and surprise surprise not a single image was badly cropped.

It could be something to do with **kindlegen**.

So I re-generated the MOBI that I generated before (the exact same settings that reproduced the issue).
And this time completely different pages were affected.

I think it's **kindlegen**.

axu2:
KCC cuts Scribe images into two sections to get around the 1920 limit of kindlegen and then reassembles it on kindle:

<img width="1123" height="630" alt="Image" src="https://github.com/user-attachments/assets/3f70bba0-10d7-4ce3-ad84-6f136e7e0b70" />

https://www.youtube.com/watch?v=OhrEWBmvhx8&t=104s

jaroslawjanas:
> That's unusual.
> 
[...]

Issue #768, opened 2024-11-03
Blank Page Issues on Kindle Colorsoft


BastiGit:
Hello,

I have had the new Kindle Colorsoft e-reader for 2 days. When I try to convert mangas with KCC using the CS12 profile, I get a white blank page every few pages. When I go back to the main menu and reopen the manga, the page then loads correctly instead of being blank.

However, this is very annoying as I have to repeat this every few pages. I didn't have this problem with my previous Kindle Paperwhite 7. Is there a workaround for this?

Thank you in advance for the response!

![grafik](https://github.com/user-attachments/assets/2ee68e61-247f-4211-90d0-e48ff872e521)

axu2:
That's concerning. Maybe the colorsoft has a bug?

Try normal jpg. 

Try no upscale.

BastiGit:
I have tried JPG, no upscale and PNG. Unfortunately, this problem also occurs here. 

However, if I turn the pages very slowly, it seems to occur only rarely.

Antheology:
> Ideas: uncheck color mode
> 
> Or
> 
> Use scribe profile with color checked.

I am also having this issue I will try these suggestions.

axu2:
@utopiafallen Do you have any insight into this issue?

utopiafallen:
> However, if I turn the pages very slowly, it seems to occur only rarely.

This has been my experience as well. Other than that, sorry, I don't have much to add.

dusk-exhale-paging:
Jumping on here to say that I am also experiencing this issue. Normally, I turn the Kindle off and back on to be able to see the pages or I get out of the book and get back in. I’ve been using KCC for a long time now with my paper white and just recently got my color soft and started to experience this blank page issue

axu2:
Can anyone test if issue happens with humble bundle epubs sent via Send to Kindle?

If it doesn't happen something is wrong with kcc epub formatting. 

If it does then it's a colorsoft bug that amazon is unlikely to fix since kcc uses the old kf8 comic format.

axu2:
I meant sending an EPUB from humble bundle directly via Send to Kindle, not using KCC at all.

axu2:
To obtain a test file, you could pay $1 for a current humble bundle:

https://www.humblebundle.com/books/pathfinder-starfinder-comics-dynamite-books

JuanPardos:
[...]

PR #1192, opened 2025-12-12
Configurable JPEG quality setting

jaroslawjanas:
This PR is here mainly to solve these issues:

- #1163
- #665 
- #768

It is not exactly clear why lower image quality (higher compression) causes these issues.
We presume that it's something to do with how the Scribe processes/renders pages?
It seems that if it takes too much time it causes issues.

Either way increasing the quality to `90` (decreasing compression) seems to have mostly fixed this problem and it's not a bad spot to be in in terms for file size.
image

Either way it is nice to have control of the quality of the images.

Changes:
  - Users can now adjust JPEG compression quality (1-100%) via a new spinbox in the GUI
  - Kindle Scribe profile now defaults to 90% quality instead of the standard 85%
  - The quality setting is automatically disabled when PNG output is selected, since PNG is lossless
  - CLI users can specify quality using the -q or --jpeg-quality flag
  - Quality preference is saved between sessions
---

Screenshots:
image
image
image
image

**AI Tool Assistance Disclosure**

This PR was developed with assistance from Claude Code using Opus 4.5. All generated code was reviewed, tested, and adjusted by a human developer.

_Don't hate the player, hate the game._

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -1181,7 +1181,7 @@ class KCCGUI(KCC_ui.Ui_mainWindow):
                 'PVOptions': True, 'ForceExpert': False, 'DefaultFormat': 0, 'DefaultUpscale': True, 'ForceColor': False, 'Label': 'KO',
             },
             "Kindle Colorsoft": {
-                'PVOptions': True, 'ForceExpert': False, 'DefaultFormat': 0, 'DefaultUpscale': True, 'ForceColor': True, 'Label': 'KCS',
+                'PVOptions': True, 'ForceExpert': False, 'DefaultFormat': 0, 'DefaultUpscale': True, 'ForceColor': True, 'Label': 'KCS', 'DefaultJpegQuality': 90,
             },
             "Kindle Paperwhite 7/10": {'PVOptions': True, 'ForceExpert': False, 'DefaultFormat': 0,
                               'DefaultUpscale': True, 'ForceColor': False, 'Label': 'KPW34'},


jaroslawjanas:
The only thing I am not sure of is if we want to pop it in and out of existence when selecting `PNG` as** we do with the **Cropping Power** slider **Preserver Margin %** when toggling `Cropping mode`.

jaroslawjanas:
I made it so that it changes visibility.
I also made the `%` in `Cropping Power` to a `QSpinBox` suffix. I think this looks better.
kcc_jpeg_quality

axu2:
This doesn't set the default on command line version. I'm not sure if that's something we should do.

I'm considering not showing JPG quality by default, since it is advanced feature. Maybe a separate checkbox.

I will take closer look soon.

jaroslawjanas:
> This doesn't set the default on command line version. I'm not sure if that's something we should do.

Good question.
I think I could make a decently clean implementation for it but the question if we should remains.

I can't answer that, I don't really use the `cli` so someone else should make the call.

Maybe a middle ground? Like we show a warning if `jpeg-quality < 90 and profile == "KS"`?
Something like
```
Warning: Kindle Scribe may have issues with JPEG quality below 90%. Use --jpeg-quality 90 to adjust.
``` 
**OR**
```
Warning: For Kindle Scribe, --jpeg-quality 90 or higher is recommended (current: {value}%).
```

jaroslawjanas:
> I'm considering not showing JPG quality by default, since it is advanced feature. Maybe a separate checkbox.

I like this, we could make a checkbox called **Advanced features**. It could house other things in the future. A nice dumping ground for all those pesky advanced configurations.

Then again is it really more *"advanced"* than **Custom gamma** or **Cropping mode**?
I won't argue about it, just sayin'.

jaroslawjanas:
So what's the plan @axu2 ?
I'd like to get this merged before I go on vacation if possible.

jaroslawjanas:
Changed the **Colorsoft** default too because of #768

axu2:
Another user reported with his Scribe 2024 that setting it to 90 fixes his problem.

Also, I needed to make quite a few changes, and it ended up being easier putting them in a separate PR, so I'll close this one. Redid commits already merged to master.

PR #1209, opened 2026-01-06
fix file order for fusion in c2e

filterpaper:
Previously, input sources were collected into a set, which caused files to be processed in an arbitrary order rather than the order specified on the command line. This change switches to using a list, ensuring that files are processed in the exact order provided by the user.


axu2:
Is options.input already a list?

filterpaper:
> Is options.input already a list?

Yes and thanks for pointing it out.

PR #1240, opened 2026-02-04
Lower minimum chunk size to 50 MB, Remarkable chunk size default of 100 MB

samsaggace:
Remarkable 2 has a hard limit to 100MB that prevents to upload bigger files.
With the chunk size to "100", the effective size is around 104MB which is too much.

To make it work for this use case:

- Set minumum to 50 (arbitrary value here)
- Add default chunk size setting to 80 (to keep a bit of room) for Remarkable 2

If it's interesting to other usecase, this new "default chunk size" setting might be used for user profiles (other Remarkable devices ?)

This is my first PR on this repo so I'm completly opened to change/improvments/comments.


axu2:
> Remarkable 2 has a hard limit to 100MB that prevents to upload bigger files.

Can you clarify what you mean by upload? Is this a web upload or a usb upload? I've never used Remarkable.

> With the chunk size to "100", the effective size is around 104MB which is too much.

Similarly, our EPUB (200 MB) sets the limit internally to 195.

> Set minumum to 50 (arbitrary value here)

That's fine

> Add default chunk size setting to 80 (to keep a bit of room) for Remarkable 2

Don't make the change in the GUI file, otherwise it is not defaulted to command line users. 95 is probably fine?

samsaggace:
Hi,

Thanks for your quick reply !

> Can you clarify what you mean by upload? Is this a web upload or a usb upload? I've never used Remarkable.

There is a lot of different options to add a file to the Remarkable (web upload, through a native app, usb with a html interface, etc.)
All of them have the same file size limit at 100MB cf [here](https://support.remarkable.com/s/article/importing-and-exporting-files#:~:text=Max%20file%20size%3A%20100%20MB) 


> Similarly, our EPUB (200 MB) sets the limit internally to 195.

Do you mean it's possible to keep 100 in the GUI but make it 95 on the CLI ?

> 
> > Set minumum to 50 (arbitrary value here)
> 
> That's fine

Ok, I'll keep that part

> 
> > Add default chunk size setting to 80 (to keep a bit of room) for Remarkable 2
> 
> Don't make the change in the GUI file, otherwise it is not defaulted to command line users. 95 is probably fine?

I havn't found somewhere else to define the profiles limits, I'll look again.
I'll do the test with 95 to see if it works.

samsaggace:
I've tester with 95MB, it's working fine!
As the limitation is for all remarkable devices, I've also set it for all profiles to 95.
On the CLI I've added a target size by default to 95 if not set, on the GUI it's also set to 95 so that user know it needs to be splitted.
If you think it's not needed on the GUI, I can remove this part

axu2:
> There is a lot of different options to add a file to the Remarkable (web upload, through a native app, usb with a html interface, etc.)
All of them have the same file size limit at 100MB cf [here](https://support.remarkable.com/s/article/importing-and-exporting-files#:~:text=Max%20file%20size%3A%20100%20MB)

A 100 MB limit on USB uploads sounds loony to me, but I guess that is what it is.

> Do you mean it's possible to keep 100 in the GUI but make it 95 on the CLI ?

It means you can keep the CLI code and leave the GUI box unchecked hopefully, always nicer to present a less cluttered UI since Rmk users don't really have a reason to change the size.

Lowering the jpg quality or using PNG with a png/lossless source may create huge space savings

> On the CLI I've added a target size by default to 95 if not set, on the GUI it's also set to 95 so that user know it needs to be splitted.
If you think it's not needed on the GUI, I can remove this part

Yea I'd prefer it removed if possible. The rc file should also be removed, those changes are machine dependent and don't matter.

axu2:
Actually, keeping the GUI change may be fine, since it's only rmk users who are more advanced, let me think about it.

the rc still needs to go though.

axu2:
Yea, I don't think defining defaults in multiple places is good, let's remove the gui changes because I doubt Rmk users will ever change it.

samsaggace:
> Yea, I don't think defining defaults in multiple places is good, let's remove the gui changes because I doubt Rmk users will ever change it.

Ok, I've removed all the GUI things related to Remarkable.
Kept the minimum to 50 though as it could be useful may be for old devices that have difficulties opening big files
CLI now set target size to 95 by default for all profiles starting with Rmk.

Tested it, works like a charm ! 👍

axu2:
Looks good and really short now, after removing the rc changes. Thanks for Remarkable support!

samsaggace:
Hi @axu2, do you think it's possible to create new release so that I can use it directly instead of my custom build ? That would be awesome, thanks

axu2:
@samsaggace Done. 9.5.0 pre-release out.

Be sure to hit the sync button the master branch on your fork on the GitHub website.

Sorry, got a bit distracted by Kindle firmware 5.19.2 on Feb 21 completely breaking KCC compatibility on Kindles.

axu2:
@samsaggace 

Hi, I just released 9.6.2, it has a lot of improvements when PNG PDF mode is selected, please try it! It should give better quality at much smaller filesize.

samsaggace:
Will test it ASAP ! Thanks again

PR #1242, opened 2026-02-08
Preserve file fusion input order with prefix

filterpaper:
When KCC is provided input files in the following order for file fusion, OS sorting will reorder the directories so that those with numeric suffixes appear first:
```
test v01.cbz
test v02.cbz
test 030.cbz
test 031.cbz
```
This change prepends a sequence index to the temporary directory names to ensure that the _user-specified order_ is preserved during the fusion process.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1559,16 +1559,14 @@ def makeFusion(sources: List[str]):
         print("Checking images...")
         source_path = Path(source)
         # Add the fusion_0001_ prefix to maintain user-specified order if needed
+        prefix = ''
         if needs_prefix:
-            if source_path.is_file():
-                targetpath = fusion_path.joinpath(f'fusion_{index:04d}_{source_path.stem}')
-            else:
-                targetpath = fusion_path.joinpath(f'fusion_{index:04d}_{source_path.name}')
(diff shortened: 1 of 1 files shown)


axu2:
I would only add the prefix if the sorted order is different. Reference older similar PR.

- https://github.com/ciromattia/kcc/pull/1010

@kiryl85 We are removing sorting in the jobList because of this reason. So people can manually specify the order. I don't think keeping it sorted is that important, right?

filterpaper:
> I would only add the prefix if the sorted order is different.

Updated the fix to support this.

filterpaper:
The prefix now uses a fixed string pattern, which is removed from chapter names, ensuring it is not visible. Let me know if this logic should still apply only when the original order differs from the sorted order.

axu2:
This adds a prefix to cbz subfolder names with cbz output, even when the original order was correct.

axu2:
Alright, looks pretty good now, made one more edit, please double check then I'll merge.

Now the overall diff looks pretty neat, even if per commit diffs are a bit complex.

filterpaper:
Thank you. I have validated the following:
* Prefixes are added only when needed and are correctly sorted in CBZ files
* Folders with prefixes are ordered correctly in EPUBs and are not displayed in TOC chapter names

PR #1276, opened 2026-03-20
if cropping, render pdf to 2x

axu2:
@9783e6 

This way, uncropped pages get a really clean 2x downscale and cropped pages don't get a slight mismatch when downscaling a small amount.


axu2:
@9783e6 

Actually this is way too expensive, I'll try a different way. Reverting.

PR #1284, opened 2026-04-07
Add kindling as preferred MOBI conversion backend

ciscoriordan:
## Summary

Adds [kindling](https://github.com/ciscoriordan/kindling) as a MOBI conversion backend, preferred over kindlegen when available.

I built kindling because kindlegen was running out of memory and taking 12+ hours trying to build Greek dictionaries with hundreds of thousands of inflected forms. It started as a dictionary-specific tool. After a user request (https://github.com/ciscoriordan/kindling/issues/1), I've since expanded it to handle EPUB-to-MOBI conversion for books and comics as well, with KF8 dual-format output, HD image container, and fixed-layout support.

Kindling is an open-source Rust replacement for kindlegen that produces the same dual KF7+KF8 MOBI output. It runs in kindlegen compatibility mode, accepting the same CLI flags (`-dont_append_source`, `-locale en`) and printing the same status codes (`:I1036:`, `:E23026:`), so the existing KCC conversion pipeline works unchanged.

Pre-built binaries for Mac (Apple Silicon + Intel), Linux, and Windows: https://github.com/ciscoriordan/kindling/releases

Tested with Pepper & Carrot comic EPUB on Kindle hardware:

Pepper & Carrot on Kindle

## Changes

- `shared.py`: New `detect_mobi_tool()` that checks for `kindling` first, falls back to `kindlegen`
- `comic2ebook.py`: `checkTools()` and `makeMOBIWorker()` use `detect_mobi_tool()` instead of hardcoded `kindlegen`
- `KCC_gui.py`: GUI detection updated, error messages mention both tools
- `Dockerfile`: Skips kindlegen symlink if kindling is already on PATH
- `README.md`: Updated installation section to recommend kindling

## Test plan

- [ ] KCC with kindling on PATH produces working MOBI from comic EPUB
- [ ] KCC without kindling falls back to kindlegen
- [ ] GUI detection works for both tools
- [ ] Docker build works with and without kindling pre-installed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,8 @@ RUN \
     ln -s /opt/kcc/kcc-c2e.py /usr/local/bin/c2e && \
     ln -s /opt/kcc/kcc-c2p.py /usr/local/bin/c2p && \
     ln -s /opt/kcc/entrypoint.sh /usr/local/bin/entrypoint && \
-    ln -s /opt/kcc/kindlegen/kindlegen /usr/local/bin/kindlegen && \
+    (test -f /usr/local/bin/kindling && true || \
+     ln -s /opt/kcc/kindlegen/kindlegen /usr/local/bin/kindlegen) && \
     cat /opt/kcc/kindlecomicconverter/__init__.py | grep version | awk '{print $3}' | sed "s/'//g" > /IMAGE_VERSION
 
 LABEL com.kcc.name="Kindle Comic Converter" \
--- a/README.md
+++ b/README.md
@@ -147,11 +147,15 @@ For flatpak, Docker, and AppImage versions, refer to the wiki: https://github.co
 
 You'll need to install various tools to access important but optional features. Close and re-open KCC to get KCC to detect them.
 
-### KindleGen
+### Kindling (recommended) or KindleGen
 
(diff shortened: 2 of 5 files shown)


axu2:
1) Looks like the attached image has formatting problems (it's not fullscreen) If you used calibre, try directly usb dropping the mobi instead not using calibre using Kindle USB file manager app or other MTP app if needed.
2) I was thinking just renaming your binary to kindlegen.exe or kindlegen and putting it on PATH or `usr/local/bin` so no code changes at all are needed, so it's truly a drop in replacement.
3) Wow, how did you add support that fast? Thought you said it would take time. You used Claude?
4) Due to Kindle update 5.19.2 completely breaking KCC compatibility, I don't plan to make any large changes like this to KCC until things calm down. Since amazon is rapidly issuing new releases rn and I don't want to introduce another variable
5) I think I'll always prefer recommending to users an official tool since the time difference isn't enormous like in your case and I'm not sure if this differs in behavior. hopefully kindlegen becomes native in macOS 28

AcidWeb:
KindleGen is main reason why I abandoned this project. It is outdated piece of shit that is abandoned by Amazon. Half of what KCC do is due to KindleGen issues or limitations.

Any realistic chance to ditch this dreadful software should be welcomed with tears of joy.
I salute you @ciscoriordan.

PR #1290, opened 2026-04-12
add webp output support (ignored for Kindle MOBI/EPUB and all PDF)

axu2:
For a manga file with all default settings:

JPG: 185 MB
mozJPG: 170 MB
PNG: 90 MB
lossy webp: 130 MN
lossless webp: 80 MB

webp doesn't support 16 color 4 bit mode in pillow... so the difference isn't too large @9783e6 

https://github.com/axu2/kcc/releases/tag/v9.6.2-webp

Because Kobo supports webp now

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,7 @@ PROCESSING:
   --coverfill           Center-crop only the cover to fill target device screen
   --forcecolor          Don't convert images to grayscale
   --forcepng            Create PNG files instead JPEG for black and white images
+  --webp                Replace JPG with lossy WEBP and PNG with lossless WEBP
   --force-png-rgb       Force color images to be saved as PNG
   --pnglegacy           Use a more compatible 8 bit PNG instead of 4 bit.
   --noquantize          Don't quantize PNG images to 16 colors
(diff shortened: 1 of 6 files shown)


axu2:
@winstonma

image

Make sure you use these settings for lossless webp. It's smaller and sharper on eink. using png mode enables lossless webp.

winstonma:
Can kindle can decode WebP image inside mobi? If yes then it would be a news to me.

axu2:
No, webp output support in kcc is just for cbz koreader. Make sure you enable lossless webp.

winstonma:
Oh I never use KCC to prepare CBZ

axu2:
@winstonma 

Here are the results of using KCC webp:

image

axu2:
> Oh I never use KCC to prepare CBZ

Would you consider using KCC in the future since it supports webp cbz output and input now? Or is there a feature missing compared to your current flow? @winstonma

winstonma:
I am using [Zealous Crop](https://docs.gimp.org/2.10/en/plug-in-zealouscrop.html) and it would remove the spaces between frames.

It makes the image 10-15% larger.

axu2:
@winstonma 

Have you tried the kcc feature "inter panel crop" which should do the same?

PR #1302, opened 2026-04-21
avoid orphan dir in tempdir and fix disk size check with tempdir

filterpaper:
* Update makeFusion to use the same temporary directory location
* Avoid creating an orphan "KCC-" in TMPDIR when "--tempdir" is set
* Ensure disk space check follows "--tempdir" setting

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -918,7 +918,7 @@ def getWorkFolder(afile, workdir=None):
                 mupdf_pdf_process_pages_parallel(afile, fullPath, target_width, target_height)
             except Exception as e:
                 rmtree(path, True)
-                raise  UserWarning(f"Failed to extract images from PDF file. {e}")
+                raise UserWarning(f"Failed to extract images from PDF file. {e}")
             return workdir
         else:
             if not os.path.exists(fullPath):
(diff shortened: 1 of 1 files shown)


kindlecomicconverter/comic2ebook.py
```diff
@@ -890,7 +894,7 @@ def getWorkFolder(afile, workdir=None):
             rmtree(workdir, True)
             raise UserWarning("Failed to prepare a workspace.")
     elif os.path.isfile(afile):
-        if disk_usage(gettempdir())[2] < os.path.getsize(afile) * 2.5:
+        if disk_usage(check_path)[2] < getDirectorySize(afile) * 2.5:
```
axu2: Why did you change `os.path.getsize()` to `getDirectorySize()`? @filterpaper 

This will always return `0` since it's a file, not a directory.

axu2 (changes requested):

axu2:
Ok, I'm going to ask to split this up, the fusion part might need more thought.

kindlecomicconverter/comic2ebook.py
filterpaper: That was an oversight on my part. Thanks for catching it.

PR #1090, opened 2025-09-20
experimental epub input

axu2:
@bust4cap @MrBlank in case you want to try, same assumptions as last time. Now i follow spine. ~If multiple images per page weird stuff might happen.~ no blank handling yet. Only extracts first image on a page.

No added dependencies this time.

Much better than 

- #1076
- fix #277 

test build: https://github.com/axu2/kcc/releases/tag/v10.1.3-epubinput

updated:

https://github.com/axu2/kcc/releases/tag/v10.1.3-1pagelandscape

PR #1342, opened 2026-05-17
epub input: extract biggest image per page

axu2:
@bust4cap

- #1090


bust4cap:
i think largest_size = 0 has to go outside the for loop, otherwise its only ever gonna pick the last image :P
otherwise its working fine :)

axu2:
Oops, wasn't paying close enough attention, my test files all had largest at the end. Thanks

PR #1355, opened 2026-05-23
File fusion uses OS temp folder if checked

### LINKED ISSUE #1350: W/B margins affect cropping
### When using intermediate W/B margins, KCC sometimes doesn't crop the picture as much as with the setting unchecked. Most of the file looks great, but for a few pages, it's not cropping nearly enough.

**Screenshots**
I think this is the best example. With margins _(unchecked)_, KCC finds the page and crops it perfectly and inserts black margins on both sides for my device's A.R.
However, without margins _(intermediate)_, KCC doesn't crop Top or Bottom nearly enough. And in this example it also didn't crop the left side.

Unchecked:
Image

Intermediate:
Image

Settings:
Image

**To Reproduce**
1. Convert two CBZs, one with intermediate W/B margins and one unchecked. Both with Cropping checked.
2. There should be a difference in cropping in some cases.

**Desktop**
 - KCC 10.1.3
 - OS: Linux Fedora 44
 - Device: Kindle Paperwhite 12th gen

**Additional context**
This seems to be a problem mostly on pages, where the margins would be black.
I've tried with Cropping Power set to 3, 2.95, 2, and 1. Same result.

filterpaper:
Use source folder when output folder is not specified.

* Fixes #1350

### DIFF
--- a/kindlecomicconverter/KCC_gui.py
+++ b/kindlecomicconverter/KCC_gui.py
@@ -393,7 +393,10 @@ class WorkerThread(QThread):
             for job in currentJobs:
                 bookDir.append(job)
             try:
+                fusion_source_parent = str(Path(bookDir[0]).parent)
                 comic2ebook.options = comic2ebook.checkOptions(copy(options))
+                if options.output is None:
+                    options.output = fusion_source_parent
                 currentJobs.clear()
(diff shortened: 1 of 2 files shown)


axu2:
Alright, simplified it. Thanks.

If you think I removed anything important let me know.

PR #1376, opened 2026-06-27
fix: enable unrar on all distros, not just Fedora

NilsLeo:
unrar was gated behind distro.id() == 'fedora', so on Debian/Ubuntu the binary was installed but never invoked. Try unrar unconditionally after 7z, with an optional rarfile Python fallback.

Fixes CBR extraction on non-Fedora Linux distributions.

### DIFF
--- a/kindlecomicconverter/comicarchive.py
+++ b/kindlecomicconverter/comicarchive.py
@@ -43,8 +43,12 @@ class ComicArchive:
     def type(self):
         extraction_commands = [
             [SEVENZIP, 'l', '-y', '-p1', self.basename],
-            ['unrar', 'l', '-y', '-p1', self.basename],
         ]
+
+        if platform.system() == 'Linux':
+            extraction_commands.append(
+                ['unrar', 'l', '-y', '-p1', self.basename],
+            )
 
         for cmd in extraction_commands:
             try:
@@ -56,13 +60,6 @@ class ComicArchive:
                 pass
             except CalledProcessError:
                 pass
-
-        try:
-            import rarfile
-            if rarfile.is_rarfile(self.filepath):
-                return 'RAR'
-        except Exception:
-            pass
 
         raise OSError(EXTRACTION_ERROR)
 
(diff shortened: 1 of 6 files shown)


axu2:
I don't want to add any more pypi dependencies.

Just change the conditional to linux, this is not needed on windows or macOS.

What error does 7z give on your file? What distro? I was under the impression that only Fedora 7z didn't support rar

NilsLeo:
The distro is Debian. I install the 7zip apt package, which ships the upstream 7zz binary, and that binary doesn't include RAR support (the RAR codec is excluded from the packaged build for licensing reasons source: https://linuxcapable.com/how-to-install-7-zip-on-debian-linux/). So on a .cbr (RAR container), 7z l/7z x fails to read it. unrar is installed, but the old code only invoked it when distro.id() == 'fedora', so on Debian it was never tried and extraction failed.



Updated per your feedback: dropped the rarfile PyPI fallback and changed the conditional to platform.system() == 'Linux' (not needed on Windows/macOS).

axu2:
Ok cool. But now I'm curious about separate issue


7zz is called on macOS. 

Kcc currently calls 7z on Linux. 

This is a separate issue, but does that mean another modification is needed? 7zz vs 7z

Also if you using an LLM please disclose which. Looks like Claude?

And do tar or bsdtar work for rar? They do on windows and macOS. But which one is called depends on OS. Kcc calls bsdtar on Linux. 

Also it'd be nice to see specific error messages and codes if possible when rar isn't supported by 7z/tar but is attempted. 

@NilsLeo

axu2:
Let me make it easier to understand. Here's is what KCC calls depending on OS

Do you see any problems with it?

| TOOL/OS | Windows | macOS | Linux  |
|---------|---------|-------|--------|
| **TAR**     | tar     | tar   | bsdtar |
| **7ZIP**   | 7z      | 7zz   | 7z     |
| **UNAR**    | n/a     | unar  | n/a    |
| **UNRAR**   | n/a     | n/a   | unrar  |

@NilsLeo

NilsLeo:
appreciate it!

axu2:
@NilsLeo Do you see any potential problems in the above table, especially regarding 7zz vs 7z or tar vs bsdtar

PR #1384, opened 2026-07-05
add Keep ComicInfo.xml in CBZ output option

### LINKED ISSUE #912: ComicInfo.xml file not moving from CBZ -> CBZ conversion
**Describe the bug✱

When creating a CBZ from a CBZ with a `ComicInfo.xml` file, the file is no longer present in the created CBZ.

**Screenshots**

Image

Image

**Desktop (please complete the following information):✱

 - OS: macOS Sequoia 15.3.1

drshliapa:
- Fixes #912

KCC currently strips ComicInfo.xml from CBZ output. This PR keeps the original
file in the output CBZ, but only for the Other profile, since ComicInfo.xml can
cause problems on Kobo (see mihonapp/mihon#664) and possibly on old Kindles, as
discussed in the issue.

Changes:
- checkOptions(): added an isOther flag that is saved before custom
  width/height overrides the profile name to "Custom". The Other profile is
  basically always used with custom dimensions, so checking
  options.profile == 'OTHER' later doesn't work.
- getMetadata(): keep the raw ComicInfo.xml content in memory instead of just
  deleting it (only for Other profile with CBZ format).
- CBZ branch of makeBook(): write the file back into the archive root before
  zipping.

Tested with the CLI:
- Other to CBZ: ComicInfo.xml is in the output, byte identical to the original
- KV to CBZ: not included
- KV with custom width/height to CBZ: not included, so the Kindle based
  "Custom" profile is not confused with Other
- Other to EPUB: no changes

Two things I wasn't sure about:
1. The file is copied as is, including the Pages element. Page indexes there
   won't match the output after renaming and spread splitting. I can strip
   Pages from the copy if that's better.
2. If you'd prefer an explicit "Keep ComicInfo.xml" checkbox (enabled only for
   the Other profile) like you suggested in the thread, I can rework it that
   way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,7 @@ OUTPUT SETTINGS:
   -t TITLE, --title TITLE
                         Comic title [Default=filename or directory name]
   --metadatatitle       Write title using ComicInfo.xml or other embedded metadata. 0: Don't use Title from metadata 1: Combine Title with default schema 2: Use Title only [Default=0]
+  --keepcomicinfo       Keep any original ComicInfo.xml files [Default=0]
   -a AUTHOR, --author AUTHOR
                         Author name [Default=KCC]
   --language            EPUB language [Default=en-US]
--- a/gui/KCC.ui
+++ b/gui/KCC.ui
@@ -22,6 +22,680 @@
     <property name="bottomMargin">
      <number>5</number>
     </property>
+    <item row="9" column="0" colspan="2">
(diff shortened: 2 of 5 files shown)


axu2:
Since my last post, I think the popularity now warrants its own option. Keeping the pages is fine, something is better than nothing and some users might not split anyways.

I want to make a new release soon, so I'll add the GUI portion myself.

PR #1385, opened 2026-07-06
Fix macOS app version showing 0.0.0

### LINKED ISSUE #632: [macOS] Kindle Comic Converter has the version number 0.0.0
starting with version 5.6.3 and including version 5.6.4 the Kindle Comic Converter claims to have version "0.0.0"

previous versions up to 5.6.2 did not have this problem, so this is a regression. starting with 5.6.3, the in your ```Info.plist``` file the ```CFBundleVersion``` was removed (which is a clear bug and should never happen) and the ```CFBundleShortVersionString``` was set to "0.0.0" (which is just nonsense)

screenshots from the Finder's GetInfo, ColumnView and QuickLook:
Screenshot 2023-11-29 at 19 13 54
Screenshot 2023-11-29 at 19 14 09
Screenshot 2023-11-29 at 19 13 37

### LINKED ISSUE #747: macOS: kkc's info.plist contains wrong version numbers since version 5.6.2
Hello!
for some time now, every macOS kkc version contains 0.0.0 as CFBundleShortVersionString - and the CFBundleVersion is missing

Screenshot 2024-09-29 at 22 05 49

drshliapa:
- Fixes #632
- fix #747 

The macOS build calls pyinstaller directly from the command line in setup.py,
so pyinstaller generates a default Info.plist with CFBundleShortVersionString
set to 0.0.0 and no CFBundleVersion at all. This is still the case in the
10.3.0 release dmg.

Following the suggestion in the issue thread to add an intermediate spec
step, this PR adds kcc-macos.spec and switches the darwin branch of
build_binary to use it. The spec was generated with pyi-makespec using the
exact same flags as the old command. The only addition is the info_plist
section in BUNDLE that reads the version from kindlecomicconverter at build
time, so nothing needs to change in the CI workflows.

Tested locally on an arm64 Mac:
- built the app with the new spec, Info.plist now shows 10.3.0 in both
  CFBundleShortVersionString and CFBundleVersion
- diffed the entire Info.plist against the official 10.3.0 dmg: the version
  fields are the only difference, everything else including CFBundleIdentifier
  is identical
- the built app launches and runs normally

I left bundle_identifier untouched on purpose to keep the change minimal for
the unsigned app.

### DIFF
--- a/kcc-macos.spec
+++ b/kcc-macos.spec
@@ -1,0 +1,61 @@
+# -*- mode: python ; coding: utf-8 -*-
+# macOS GUI build, used by setup.py build_binary
+
+import sys
+
+sys.path.insert(0, SPECPATH)
+from kindlecomicconverter import __version__
+
+
+a = Analysis(
+    ['kcc.py'],
+    pathex=[],
+    binaries=[],
+    datas=[],
+    hiddenimports=['_cffi_backend'],
+    hookspath=[],
+    hooksconfig={},
+    runtime_hooks=[],
+    excludes=[],
+    noarchive=False,
+    optimize=0,
+)
+pyz = PYZ(a.pure)
+
+exe = EXE(
+    pyz,
+    a.scripts,
+    [],
+    exclude_binaries=True,
+    name='Kindle Comic Converter',
+    debug=False,
+    bootloader_ignore_signals=False,
+    strip=True,
+    upx=True,
+    console=False,
+    disable_windowed_traceback=False,
+    argv_emulation=False,
+    target_arch=None,
+    codesign_identity=None,
+    entitlements_file=None,
+    icon=['icons/comic2ebook.icns'],
+)
+coll = COLLECT(
+    exe,
+    a.binaries,
+    a.datas,
+    strip=True,
+    upx=True,
+    upx_exclude=[],
+    name='Kindle Comic Converter',
+)
+app = BUNDLE(
+    coll,
+    name='Kindle Comic Converter.app',
+    icon='icons/comic2ebook.icns',
+    bundle_identifier=None,
+    info_plist={
+        'CFBundleShortVersionString': __version__,
+        'CFBundleVersion': __version__,
+    },
+)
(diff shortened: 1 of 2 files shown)


drshliapa:
image
image

axu2:
Thanks, LGTM. I see it's based on this section of pyinstaller docs: https://pyinstaller.org/en/stable/spec-files.html#using-shared-code-and-configuration-in-spec-files

verified that the generated spec file is identical besides the version data.

Please disclose any LLM usage, if any.

edited your first post so the issues it fixes are expanded in title

I'm still thinking about your other PR, I'll get to it.

drshliapa:
Yes, Claude Code. I use it for research and drafting, I'm moving to Python
from Ruby and iOS development, so I lean on it for the less familiar parts
like pyinstaller. Same applies to #1384 and the issue comments. The changes
were reviewed and verified by me locally: built the app, diffed the resulting
Info.plist against the official 10.3.0 dmg, the screenshots above are mine.

And yes, the SPECPATH import follows that section of the docs. One thing I
did differently on purpose: the BUNDLE version argument only sets
CFBundleShortVersionString, while the report in #632 was also about
CFBundleVersion missing entirely, so I set both keys through info_plist
instead.

axu2:
On second thought, would it be better to instead:

> by passing version string argument to the BUNDLE.

which is part of 

> The extra entries for generated Info.plist can already be specified in the .spec file, by passing info_plist dict to the BUNDLE. So CFBundleShortVersionString can be passed either via that dict, or by passing version string argument to the BUNDLE.

https://github.com/pyinstaller/pyinstaller/issues/6943#issuecomment-1183594834
https://pyinstaller.org/en/stable/spec-files.html#spec-file-options-for-a-macos-bundle

drshliapa:
I think you're right, this is cleaner.

The current version works because `info_plist` can set both values, but using `version=__version__` for `CFBundleShortVersionString` is more explicit. `CFBundleVersion` still needs to stay in `info_plist`, since PyInstaller's `version` argument only sets `CFBundleShortVersionString`.

So I'd change it to:

```python
app = BUNDLE(
    coll,
    name='Kindle Comic Converter.app',
    icon='icons/comic2ebook.icns',
    bundle_identifier=None,
    version=__version__,
    info_plist={
        'CFBundleVersion': __version__,
    },
)
```

axu2:
I see, it only sets one. In that case, let's just leave it as is.

PR #1397, opened 2026-07-19
Bound multiprocessing pools by KCC_WORKERS to cap memory

NilsLeo:
The PDF render/extract pool, the image-processing pool (`imgDirectoryProcessing`, which runs for every conversion) and the MOBI build pool each spawn `cpu_count()` workers. Each worker holds a full-resolution page or image in memory, so peak memory scales with the core count and can exhaust RAM on large or high-DPI files — a real problem on memory-limited/containerised hosts (an 8-core box can need ~8× the per-page memory).

This adds `pdf_render_worker_count()`, which honours a `KCC_WORKERS` environment variable as an upper bound and falls back to `cpu_count()` when unset, and applies it to all three pools. The MOBI pool previously sized itself from `virtual_memory().total`, which reports host RAM rather than a container's memory limit, so it is now also capped.

Default behaviour is unchanged when `KCC_WORKERS` is unset.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -678,7 +678,9 @@ def buildPDF(path, title, job_progress='', cover=None, output_file=None):
 
 def imgDirectoryProcessing(path, job_progress=''):
     global workerPool, workerOutput
-    workerPool = Pool(maxtasksperchild=100)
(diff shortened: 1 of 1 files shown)


axu2:
Ah, sorry I didn't take a look quite yet, is this PR not needed? Did you find an alternate solution?

PR #1398, opened 2026-07-19
Skip unreadable pages during PDF extraction instead of failing the whole book

NilsLeo:
When PyMuPDF raises while rendering or extracting a single page, the exception propagates out of the worker `Pool` and often cannot be pickled (`SwigPyObject`), so the entire job dies with an opaque `TypeError` about pickling instead of the real cause. A PDF with a handful of bad pages fails completely even though every other page is fine.

This catches per-page errors inside the worker, logs a warning, skips the offending page and continues — so the book still converts, and genuine failures surface with a clear message rather than a masked pickling error.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -803,16 +803,22 @@ def render_page(vector):
         seg_to = min(seg_from + seg_size, num_pages)  # last page number
 
         for i in range(seg_from, seg_to):  # work through our page segment
-            page = doc[i]
-            if not pdf_width or page.rect.width > page.rect.height:
-                zoom = target_height / page.rect.height
-            else:
-                zoom = target_width / page.rect.width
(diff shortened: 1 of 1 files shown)


axu2:
Do you have a small sample example PDF file for failing rendering and any other relevant examples for other PRs?

NilsLeo:
Yes, quite busy atm, but will try to provide relevant examples for each bug ie the applied settings and the file which triggered the exception

axu2:
I guess I'm more interested in the exact underlying error messages (sometimes kcc hides the underlying error since error handling isn't the best)

axu2:
Just to be clear, you had errors in both render and extract cases? Only 1 is used per pdf. So you'd have two example files.

NilsLeo:
Yes, unfortunately I haven't gotten this error in a while. Might've been a real outlier so I'll close it and reopen if it ever shows again :)

PR #1399, opened 2026-07-19
Skip corrupt pages instead of aborting the whole conversion

NilsLeo:
A single unreadable or zero-byte page raised `RuntimeError` and failed the entire book, even when every other page was fine. This skips the bad page and carries on, while keeping a genuine environment error (a missing Pillow decoder) fatal so real setup problems still surface. It also guards against a missing cover, so a damaged archive with no usable cover page no longer crashes with a `NoneType` error.

## How to reproduce / test

A comic with one **zero-byte** page and one **unreadable/garbage** page mixed in with valid pages triggers both fatal branches this PR fixes:

**Download:** https://drive.google.com/file/d/1rz38SkzJLcZ3D_laJCdC8XYo5nRg9Ghi/view?usp=drive_link

Save it next to where you run the commands (referred to below as `1399.cbz`). The archive is flat (images at the root) with `page-003-EMPTY.jpg` (0 bytes) and `page-005-GARBAGE.jpg` (non-image bytes) between valid pages.

```bash
```

**1. Reproduce the crash on upstream master** (one corrupt page fails the whole book):
```bash
kcc ciromattia/kcc:master -p KV -f EPUB -o out 1399.cbz
```
→ aborts with `RuntimeError: Image file .../page-005-GARBAGE.jpg is corrupted.` at `detectSuboptimalProcessing`; no EPUB produced.

**2. Same file on this branch — skips the bad pages and converts the rest:**
```bash
kcc NilsLeo/kcc:up/image-resilience -p KV -f EPUB -o out 1399.cbz
```
→ logs
```
WARNING: Skipping corrupt image .../page-005-GARBAGE.jpg (cannot identify image file ...)
WARNING: Skipping corrupt (empty) image .../page-003-EMPTY.jpg
```
then writes a valid `out/1399.epub` from the remaining pages.

(Or clone this branch and run `python kcc-c2e.py -p KV -f EPUB -o out 1399.cbz` directly.)

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1368,8 +1368,11 @@ def detectSuboptimalProcessing(tmppath, orgpath):
                 path = os.path.join(root, name)
                 pathOrg = orgpath + path.split('OEBPS' + os.path.sep + 'Images')[1]
                 if os.path.getsize(path) == 0:
-                    rmtree(os.path.join(tmppath, '..', '..'), True)
-                    raise RuntimeError('Image file %s is corrupted.' % pathOrg)
+                    # A single corrupt (zero-byte) page should not fail the whole book;
+                    # skip it and continue with the remaining pages.
+                    os.remove(path)
+                    print('WARNING: Skipping corrupt (empty) image %s' % pathOrg)
+                    continue
                 try:
                     img = Image.open(path)
                     imageNumber += 1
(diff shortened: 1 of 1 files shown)


NilsLeo:
@axu2 this is constructed, but ive encountered files like this several times that have these sort of corrupted pages

axu2:
Thanks, I'm still thinking about whether or not kcc should abort if there are any corrupted pages. I feel a user should delete corrupted pages manually in that case if that is truly the behavior they want so they 100% know what pages are being skipped.

I've seen the cannot identify image file appear when the extractor is outdated. Maybe try different extractors. For example, on macOS 7zz vs 7z or unar.

edit: maybe a command line arg to skip corrupt...

NilsLeo:
Good idea - or maybe a warning text which prints skipped pages?

PR #1403, opened 2026-07-20
comic2panel: fall back to per-page splitting when merged strip is too tall

NilsLeo:
## Problem

When panel-split mode (`-i`) is active, `mergeDirectory` stacks all images in a chapter folder into a single tall strip before detecting panel boundaries. If the combined height exceeds `131072 × 4 = 524 288 px` (GIMP_MAX_IMAGE_SIZE), a `RuntimeError` is raised inside the multiprocessing worker.

The exception is caught by the `except Exception` block and returned as an error tuple. `mergeDirectoryTick` sees the non-`None` output, terminates the pool, and the entire job fails with:

```
RuntimeError: One of workers crashed. Cause: Image too tall at N pixels. …
```

The comment above the check already says *"Silently drop directories that contain too many images"* — but the code raises instead of dropping.

## Fix

Return early (`return`) instead of raising. `mergeDirectoryTick` receives `None` (falsy), so no error is recorded and the pool runs to completion. The original page images stay on disk and are processed by the subsequent `splitImage` pass, which panel-splits each page individually.

Per-page splitting produces correct panel detection for every page; the only thing lost is cross-page panel merging for that oversized directory, which wasn't working anyway.

## Testing

Reproduce with a chapter containing enough tall pages to exceed ~524 k combined pixels. Before this change the job errors. After it completes with per-page panel splits.

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1357,6 +1357,13 @@ def detectSuboptimalProcessing(tmppath, orgpath):
                     raise RuntimeError('Image file %s is corrupted.' % pathOrg)
                 try:
                     img = Image.open(path)
+                    if 0 in img.size:
+                        # Zero-dimension image (broken/blank page): skip it instead of
+                        # crashing later in resize/thumbnail with ZeroDivisionError.
+                        img.close()
+                        os.remove(path)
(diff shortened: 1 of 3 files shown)


NilsLeo:
Opened by mistake — intended for a private fork.

PR #1406, opened 2026-07-24
Fix crash when an archive wraps all pages in a single folder

NilsLeo:
When an archive (CBZ/CBR/ZIP) contains all its images inside **one top-level folder**, `getWorkFolder` flattens them by moving the files up into a fresh work directory — but the destination `OEBPS/Images` folder was never created (`mkdtemp` only makes the temp root). `shutil.move` then raises `FileNotFoundError: .../OEBPS/Images` and the entire conversion fails, even though every page is valid.

This creates the destination directory before the move.

```
  File ".../comic2ebook.py", line ~1009, in getWorkFolder
    move(os.path.join(fullPath, tdir[0], file), fullPath2)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/KCC-xxxx/OEBPS/Images'
```

## Reproduce / test

Any archive whose pages all live inside a single wrapping folder, e.g. `MyComic/MyComic - 001.jpg`, triggers it. The easiest way to A/B any branch without a local setup is [`kcc-run`](https://github.com/NilsLeo/kcc-run), which clones a given `owner/repo:ref` into a throwaway container at runtime:

```bash

# master — aborts in getWorkFolder with FileNotFoundError, no output
kcc ciromattia/kcc:master        -p KV -f EPUB -o out wrapped.cbz

# this branch — flattens the folder and converts the rest
kcc NilsLeo/kcc:up/subfolder-flatten -p KV -f EPUB -o out wrapped.cbz
```

### DIFF
--- a/kindlecomicconverter/comic2ebook.py
+++ b/kindlecomicconverter/comic2ebook.py
@@ -1357,6 +1357,13 @@ def detectSuboptimalProcessing(tmppath, orgpath):
                     raise RuntimeError('Image file %s is corrupted.' % pathOrg)
                 try:
                     img = Image.open(path)
+                    if 0 in img.size:
+                        # Zero-dimension image (broken/blank page): skip it instead of
+                        # crashing later in resize/thumbnail with ZeroDivisionError.
+                        img.close()
+                        os.remove(path)
+                        print('WARNING: Skipping zero-dimension image %s' % pathOrg)
+                        continue
                     imageNumber += 1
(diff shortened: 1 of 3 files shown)


NilsLeo:
Closing — upstream master already contains this exact fix (os.makedirs(fullPath2, exist_ok=True) before the single-subfolder flatten move). Our fork had simply fallen behind on this line; fixed on the fork side. Thanks!

Issue #526, opened 2023-05-27
[Feature request] Option to select exceptions for the 'cropping mode' feature


distample:
Sometimes there is a need for some pages in the book not to crop the margins, but now the checkbox affects all pages without exceptions. 

Could you please add functionality where pages are added to exceptions before processing and they are not cropped?

axu2:
Can you add details on which Kindle/resolution you are using and what resolution the source images are?

distample:
Maybe something like a structure scanner that outputs it to the program window, and there it marks which pages DO NOT need to be cropped?

```
import os
import zipfile
from bs4 import BeautifulSoup


def extract_cbz_structure(cbz_path, extract_dir):
    with zipfile.ZipFile(cbz_path, 'r') as cbz_file:
        cbz_file.extractall(extract_dir)


def generate_html_with_checkboxes(extract_dir):
    html = "<html><body>"
    for root, dirs, files in os.walk(extract_dir):
        relative_path = os.path.relpath(root, extract_dir)
        indent_level = relative_path.count(os.sep)
        indent = "&nbsp;&nbsp;&nbsp;&nbsp;" * indent_level
        html += f'{indent}<input type="checkbox" name="{root}" value="{root}">{relative_path}<br>'
        for file in files:
            if file == 'ComicInfo.xml':
                continue  # Skip the ComicInfo.xml file
            file_path = os.path.join(root, file)
            relative_file_path = os.path.relpath(file_path, extract_dir)
            indent_level_file = relative_file_path.count(os.sep)
            indent_file = "&nbsp;&nbsp;&nbsp;&nbsp;" * (indent_level_file + 1)
            html += f'{indent_file}<input type="checkbox" name="{file_path}" value="{file_path}">{file}<br>'
    html += "</body></html>"
    return html


# Path to your cbz file
cbz_path = "/path/to/file.cbz"
# Folder to extract the cbz file structure
extract_dir = "/path/to/folder"

# Extract the cbz file structure
extract_cbz_structure(cbz_path, extract_dir)

# Generate HTML with checkboxes
html = generate_html_with_checkboxes(extract_dir)

# Write HTML to file
with `open("/path/to/output.html",` "w") as f:
    f.write(html)
```

PR #1441, opened 2026-09-16
don't crop first page if its color

axu2:
@neyney10 just FYI

https://github.com/axu2/kcc/releases/tag/v11.3.0-dont-crop-cover


axu2:
@neyney10 This is mostly because people were complaining about covers with solid yellow backgrounds getting cropped. They didn't want to crop the yellow background, which mapped to value 212 in grayscale.

axu2:
related?

- #526 

@distample

